PR #5, opened 2015-01-25
removed data.type from widgets, it is set automatically in create_pages

peuter:
in addition to our discussion in #3


src/structure/pure/break.js
```diff
@@ -22,7 +22,6 @@ design.basicdesign.addCreator('break', {
   create: function( element, path, flavour, type ) {
     var data = templateEngine.widgetDataInsert( path, {
       path: path,
```
ChristianMayer: Won't this break the code (trailing comma)?

PR #148, opened 2015-04-02
new client for openHAB2 + infoaction plugin + Bugfixes

peuter:
- new SSE-based client for openHAB2
- new infoaction plugin (combines an action-widget with an info-widget, inspired by openHAB´s group items)
- additional info-widget in a pagejump (e.g show number of open windows in the pagejump and jump to the page where all windows are shown in detail)
- bugfixes for colorchooser in rgb mode (for openHAB color items)

### DIFF
--- a/src/lib/cometvisu-client-openhab.js
+++ b/src/lib/cometvisu-client-openhab.js
@@ -1,4 +1,4 @@
-/* cometvisu.js (c) 2010 by Christian Mayer [CometVisu at ChristianMayer dot de]
+/* cometvisu-client-openhab.js (c) 2015 by Tobias Bräutigam [tbraeutigam at gmail dot com]
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the Free
--- a/src/lib/cometvisu-client-openhab.js
+++ b/src/lib/cometvisu-client-openhab.js
@@ -56,6 +56,7 @@ function CometVisuOh( urlPrefix )
   this.handleMessage = function (e) {
 	  var json = JSON.parse(e.data);
 	  var data = json.d;
+	  console.log(data);
 	  thisVisu.lastIndex = e.lastEventId;
 	  thisVisu.update( data );
   }
--- a/src/lib/cometvisu-client-openhab.js
+++ b/src/lib/cometvisu-client-openhab.js
@@ -1,0 +1,103 @@
+/* cometvisu.js (c) 2010 by Christian Mayer [CometVisu at ChristianMayer dot de]
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 3 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+ */
+
+/**
+ * The JavaScript library that implements the CometVisu protocol.
+ * @module CometVisu Client
+ * @title  CometVisu Client
+ * @requires jQuery
+*/
+
+/**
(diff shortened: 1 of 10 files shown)


ChristianMayer:
Great to see continued OpenHAB support!

src/lib/cometvisu-client-openhab.js
```diff
@@ -0,0 +1,112 @@
+/* cometvisu.js (c) 2010 by Christian Mayer [CometVisu at ChristianMayer dot de]
```
ChristianMayer: Can you explain a bit about the new protocoll and why it needs a new communication lib / can't be using a unified communication lib?

And as it's "your" file you can place yourself in the first line, of course :)

src/lib/templateengine.js
```diff
@@ -1321,6 +1336,8 @@ function TemplateEngine( undefined ) {
     
     if (page_id.match(/^id_[0-9_]*$/) == null) {
       // find Page-ID by name
+      // decode html code (e.g. like &apos; => ')
+      page_id = $("<textarea/>").html(page_id).val();
```
ChristianMayer: I'm not too fond of using JQuery (carelessly) anymore as it's somtimes quite slow... It's good to get code written quickly - but not for quick code ;)

In this specific case: do we need such an universal solution? Or would a more specific escape / mapping also work?

src/plugins/infoaction/structure_plugin.js
```diff
@@ -0,0 +1,85 @@
+/* structure_plugin.js (c) 2014 by Michael Hausl [michael at hausl dot com]
```
ChristianMayer: Can you explain a bit more what this plugin will do? (E.g. in the forum as I guess others might want to use it as well)

src/structure/pure/pagejump.js
```diff
@@ -48,7 +47,15 @@ design.basicdesign.addCreator('pagejump', {
       'align'   : $e.attr('align'),
       'target'  : target
     } );
-    return ret_val + actor + '</div>';
+    var info = '';
+    var infoWidget = $('infowidget > *', $e).first()[0];
```
ChristianMayer: Can this be solved differently as a "dependancy" of a widget of a totally different plugin usually isn't good architecture...

src/lib/cometvisu-client-openhab.js
peuter: It uses Server-side-events (SSE), which is basically a long-polling where the connection is not closed after data was send (so no reconnection overhead). Please take a look at: https://jersey.java.net/documentation/latest/sse.html. The new openHAB2 does not support long-polling anymore, so we need a new communication lib. And IMHO SSE is better than long-polling. I can already see the performance benefit of gettnig rid of the reconnection overhead, although this is an early version (of course this only recognizable, when you have lots of updates in your visu).
Another advantage is, that the browser cares about lost connections and reconnecting.
So if you want to unify the communication I would rather suggest to update the default backend to use SSE as well.

I think the code is much cleaner this way, than putting it all in one lib.

src/lib/templateengine.js
peuter: I stumbled upon the problem that a pagejump named "Child's room" didn´t work. In the config it must be written like "Child&apos;s Room" and because of that the scrollToPage function does not find the matching page.
There is not html_(de|en)code functions in javascript, and after some googling the jquery-way was the only solution i found. So you need a complete html escaping method here to make sure everything works. Any suggestions?

src/structure/pure/pagejump.js
peuter: it is not dependant from the infoaction plugin, but uses the info-widget, which is, like the pagejump itself, part of the pure-structure. IMHO this is no bad architecture, because they "live in the same namespace". The other solution would be to copy the whole code from info.js to pagejump.js, and that would be bad architecture/coding e.g. redundant code.

src/plugins/infoaction/structure_plugin.js
peuter: Done: http://knx-user-forum.de/forum/supportforen/cometvisu/824414-vorstellung-infoaction-plugin

src/structure/pure/pagejump.js
ChristianMayer: Ah, you are right - the night was too late and I misread info as infoaction :)

src/lib/cometvisu-client-openhab.js
ChristianMayer: I really think we should unify that code - but at the moment I don't know how, though...

So I'll merge it now and we can try to make a unification later

PR #156, opened 2015-04-09
- added default mapping to metal-config

### LINKED ISSUE #153: No default icon without data from the bus/cache
For further information see here:

http://knx-user-forum.de/forum/supportforen/cometvisu/825439-default-icon-aus-mapping-ohne-aktuellen-wert-nicht-angezeigt-in-aktueller-git-version#post825439

peuter:
- fix for default value handling (fixes #153)

I´m not sure if this completely fixes #153. Are there any widgets that need to call the defaultUpdate method and do not use the createDefaultWidget-method?

### DIFF
--- a/src/config/demo/visu_config_metal.xml
+++ b/src/config/demo/visu_config_metal.xml
@@ -19,7 +19,7 @@
         </entry>
       </mapping>
       <mapping name="OnOff">
-        <entry value="0">O</entry>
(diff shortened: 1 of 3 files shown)


ChristianMayer:
"Are there any widgets that need to call the defaultUpdate method and do not use the createDefaultWidget-method?" - I don't know by heart, but it is allowed to do so

PR #167, opened 2015-04-22
some bugfixes for openHAB2

peuter:
- correct usage of requirejs-dependency of openhab-client
- fixes for editor usage in openHAB2
- code cleanup


src/editor/index.php
```diff
@@ -118,7 +132,8 @@ function simplifyPath( $path ) {
 }
 
 // .. as a fully qualified filename
-$strSchemaFQFilename = realpath( simplifyPath( $strConfigPath . ($isDemo?'demo/':'') . $strSchemaFilename) );
+// $strSchemaFQFilename = realpath( simplifyPath( $strConfigPath . ($isDemo?'demo/':'') . $strSchemaFilename) );
+$strSchemaFQFilename = realpath($strConfigPath . ($isDemo?'demo/':'') . $strSchemaFilename );
```
ChristianMayer: Why is the simplifyPath missing here? This might cause problems when the config files are located externally (e.g. /etc/cometvisu/ and only linked into the web root)

peuter: What exactly is simplyfyPath doing  that isn't already covered by realpath ?

ChristianMayer: realpath works on filesystem paths, simplifyPath only on the string that represents a path.

When it comes to symlinks it makes a huge difference. This is relevant when the web root contains symlinks as then the ".." in a URL will collide with a ".." in the file system.
Leaving it out will cause file-not-found as a best and a security leak as a worst case.

peuter: Can you provide a simple example that reproduces the problem (I admit that I didn´t understand the problem yet: Is simplifyPath something like realpath, but does not allow to "breakout" the config folder?), because simplifyPath isn´t working in openHAB2 environment. And so i have to find another solution than just removing it.

ChristianMayer: See the thread http://knx-user-forum.de/forum/supportforen/cometvisu/41563-visu_config-xsd-existiert-nicht 

The problem happens when /etc/cometvisu gets symlinked as .../cometvisu/config and the XSD has it's normal place outside .../cometvisu/config. Now the link in the XML refers to it by ".." which is fine when you call the string manipulation of the URL and then switch to the file system. But when you switch to the file system fist and then apply the .. there you'll end in a different directory.

peuter: Ok, it was a windows slash vs. backslash problem. simplifyPath re-enabled

PR #170, opened 2015-04-24
Fix pagejump active state

joltcoke:
Currently the active state doesn't get updated, if you have a nested navbar


ChristianMayer:
Are you sure that this doesn't create a regression?

peuter:
I think this PR solves the problem at the wrong end. If you take a closer look at the widgetData structure, page-widgets are the only widgets that store their data in path+'_' and not just path. Is there any reason for that? If you change that this fix would be obsolete.

ChristianMayer:
@peuter was that comment for me or for @joltcoke ?

About the IDs/paths: an path ending with '_' is the page itself, one ending with a number is a widget - like the page-widget.
So this is necessary to distinguish  between a page-widget and the page itself.

peuter:
@ChristianMayer Ok thanks for the explanation.
@joltcoke Could you delete the commented line https://github.com/joltcoke/CometVisu/blob/develop/src/structure/pure/pagejump.js#L78 as it is an leftover of older fix for the same issue.
After that is done, this PR should be merged.

joltcoke:
@ChristianMayer No, i'm not sure it is creating a regression, i've done some testing to verify i'm not breaking something else, but without proper test cases i'm not able to guarantee it.

Active state handling was completely broken, your commit 291a17c234a47fd4ea861a0bac4d1d94b33ac202 fixed 50% (only top level navbar) of my problems. I've applied the same fix to a different location to fix the rest of my problems (nested navbars).

I've tested the metal demo, it is still broken (and was broken before my fix). I'll need to dig deeper and will update the PR later.

joltcoke:
Here is my analysis:

Basically the structure of the metal demo looks like this:

```
00: <page Übersicht>
01:   <navbar>
02:     <pagejump Übersicht>
03:     </pagejump>
04:     <pagejump Wohnzimmer>
05:     </pagejump>
06:   </navbar>
07:   <group Allgemein>
08:   </group>
09:   <page KNX>
10:     <navbar>
11:       <pagejump Wohnzimmer>
12:       </pagejump>
13:       <pagejump Schlafzimmer>
14:       </pagejump>
15:     </navbar>
16:     <page Wohnzimmer>
17:     </page>
18:     <page Schlafzimmer>
19:     </page>
20:   </page>
21: </page>
```

The pagejump in line 04 is targeting a sub page of a page. Is that a valid behaviour? This is causing a problem when finding the active state. When starting the search with page Schlafzimmer selected, we will never hit the target (Wohnzimmer) of the top level pagejump because it is on the same level as the page we started the search on.

To me the proper fix would be:

- Convert page Wohnzimmer (line 16+17) into a group
- Have the top level navbar target KNX instead of Wohnzimmer

Alternatively we could also adjust the active state search algorithm to dive into all child pages to find potential jump targets.

Opinions?

ChristianMayer:
Targeting a sub-page is perfectly valid.

ChristianMayer:
Playing a bit with the code I tend to think that this commit is fine and not containing a regression.

peuter:
@joltcoke My suggestion is to change the top-level navbar target to KNX

joltcoke:
@peuter You need to convert page Wohnzimmer into a group, otherwise the behavior will change (CV will generate a landing page listing the names of the sub pages.

I'm currently looking into a fix for sub page targets.

peuter:
@joltcoke I know, but as this is "just" a demo config, I´m fine with this change of behaviour.

joltcoke:
Supporting sub page targets seems to be difficult. As there are other showstoppers to be fixed, i'd rather keep it the way it is in this PR. In most cases it will work just fine, some cases (sub page targets) are currently broken but there is a workaround.

@ChristianMayer do you agree?

ChristianMayer:
I think subpages should be supported. Just imagine a house based page structure like:
- ground floor
 - living room
 - kitchen
 - floor
- upper floor
 - Bath
 - Kids
 - ...

Why should a pagejump not offer direct access to e.g. the living room?

But currently I can't see where this would not work at the moment.

What is a problem though is when a page name is shared at different places of the hierarchy. This can be seen in the demo config: Flavour Test -> Flavour: * -> Subpage

peuter:
The pagejump to a subpage is working fine, but it isn´t shown as active ancestor. Let´s stick to your example: if you have a topnavbar with the "floor-pagejumps" and e.g. ground floor jumps directly to living room, the ground floor pagejump is inactive when you jump to e.g. kitchen.
To cover this case we must add something to the configuration where you can define an additional subpage-target.

ChristianMayer:
Ah, I think I understand that problem now.

Probably we should extend the syntax by a path / ancestor information. This can solve the issue here as well as link to the correct "toilet" in the example above.

So I suggest to merge this request and open a new issue about the "lost ancestors" (with a link to this discussion... :)) - or even fix it with a new pull request.

peuter:
Ok, merge this one and continue discussion in #177

PR #176, opened 2015-04-26
Popup event

ChristianMayer:
This will call a "close" event when a popup is being clicked.
This close event can be also called externally to close a popup.
Or you can listen externally to a close event for your own actions at that time. E.g. a diagram could listen to it to stop data fetching.

(cf. http://knx-user-forum.de/forum/supportforen/cometvisu/828864-diagramme-werden-permanent-aktualisiert )


ChristianMayer:
Didn't want to close it. It's still a valid pull request.

PR #177, opened 2015-04-27
quick solution proposal for subpage-pagejumps

peuter:
continue #170 

this is just meant as a discussion proposal, to find the best solution (although its working fine)
@joltcoke @ChristianMayer Let me know what you think.

### DIFF
--- a/src/lib/templateengine.js
+++ b/src/lib/templateengine.js
@@ -1338,22 +1338,77 @@ function TemplateEngine( undefined ) {
       + '" id="'+path+'" data-type="'+data.type+'"/>').append(retval);
     }
   };
-
-  this.scrollToPage = function(page_id, speed, skipHistory) {
-    if( undefined === page_id )
-      page_id = this.screensave_page;
-    
-    if (page_id.match(/^id_[0-9_]*$/) == null) {
+  
+  this.getPageIdByPath = function(page_name, path) {
+    if (page_name==null) return null;
+    if (page_name.match(/^id_[0-9_]*$/) != null) {
+      // already a page_id
+      return page_name;
+    } else {
+      if (path!=undefined) {
+        var scope = templateEngine.traversePath(path);
+        if (scope==null) {
+          // path is wrong
+          console.error("path '"+path+"' could not be traversed, no page found");
+          return null;
+        }
+        return templateEngine.getPageIdByName(page_name,scope);
+      } else {
+        return templateEngine.getPageIdByName(page_name);
+      }
+    }
+  }
+  
+  this.traversePath = function(path,root_page_id) {
+    var path_scope=null;
+    var index = path.indexOf("/");
+    if (index>=1) {
+      // skip escaped slashes like \/
+      while (path.substr(index-1,1)=="\\") {
+        var next = path.indexOf("/",index+1);
+        if (next>=0) {
+          index=next;
+        }
+      }
+    }
+//    console.log("traversePath("+path+","+root_page_id+")");
+    if (index>=0) {
+      // traverse path one level down
+      var path_page_name = path.substr(0,index);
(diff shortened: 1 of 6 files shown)


src/lib/templateengine.js
```diff
@@ -1394,6 +1391,14 @@ function TemplateEngine( undefined ) {
         });
       }
     }
+    return page_id;
+  }
+
+  this.scrollToPage = function(page_id, speed, skipHistory) {
+    if( undefined === page_id )
+      page_id = this.screensave_page;
+    page_id = thisTemplateEngine.getPageIdByName(page_id);
```
ChristianMayer: When the parameter of scrollToPage is not the page_id I would rename it to avoid confusion

ChristianMayer:
Would this solve the link to a specific "toilet" in the example:
- ground floor
 - living room
 - kitchen
 - floor
 - toilet
- upper floor
 - bath
 - kids
 - toilet
 - ...

peuter:
Well somehow but not really. You could use two pagejumps like
<pagejump target="ground floor" child="4">
<pagejump target="upper floor" child=3">
But just would be an unsufficient workaround. If we really want to fix this we should introduce paths to the target like:
<pagejump target="ground floor/toilet">
<pagejump target="upper floor/toilet">

ChristianMayer:
I think paths is the better way to go.

Right now you could already use the ID - but users didn't like it for good reasons. Child-numbers would be similar, I guess.

What about a new, optional attribute that contains the path?

One thing to consider when you use paths: the path separator might also also be used in the page names (especially the obvious characters like ' ' or '/'...).

peuter:
I think you are mixing up the two problems:
1. "same pagename in different paths"
2. "ancestor-pagejump"

If we have this pagejump in our topnavbar:
```
<pagejump target="ground floor/living room">
```
we still need a way to define that this pagejump is active on any "ground floor"-page (problem 2). Introducing paths only solves problem 1

btw. what path separator do you suggest we could use here?

peuter:
Ok new proposal:
```
<pagejump path="KNX" target="Wohnzimmer" active_scope="path">
```
path addresses problem 1, active_scope addresses problem 2 (where active_scope="target" is the default behaviour).
As "path" is just an ancestor and no real path, I´m not really happy with the name. Would an ancestor be enough to address problem 1 or do we need a "real" path here?

ChristianMayer:
I think we should solve 1 + 2 at the same time as we are at it at the moment.
Wouldn't 2 be solved automatically when the path/target is unique? Eveny pagejump that has a path contained in the target-path could be marked active, IMHO.

For a good path separator I have no clue as all obvious possibilities are also very likely to be used in a normal name as well. Perhaps we could use escaping? (e.g. `/` = separator, `\/` = `/` and `\\` = `\`)

peuter:
My last proposal solves both problems. Just using path/target is not enough.
```
<pagejump path="ground floor" target="living room">
<pagejump path="ground floor" target="kitchen">
<pagejump path="ground floor" target="toilet">
<pagejump path="ground floor" target="floor">
```
I need a way to choose one of them to be active on all "ground floor" pages.

```
<pagejump path="ground floor" target="living room" active_scope="path">
<pagejump path="ground floor" target="kitchen">
<pagejump path="ground floor" target="toilet">
<pagejump path="ground floor" target="floor">
```
I don´t see another way to solve this. Please provide an example, how to solve this, by just using path/target. If we mark all pagejumps as active, which have a path contained in the target path all 4 pagejumps would be active as long as one of the pages is active, thats not what we want.

peuter:
I added path handling where ground floor/toilet can be used and escaped slashes are ignored. From my point of view this would be a sufficient solution and I propose to merge this.

PR #178, opened 2015-05-02
Switch CometVisu to use "use strict" mode

ChristianMayer:
This helps avoid mistakes, etc. pp.


src/lib/cometvisu-client-openhab.js
```diff
@@ -22,6 +22,9 @@
  * @requires jQuery
  */
 
+define( ['cometvisu-client'], function( CometVisu ) {
+  "use strict";
+
```
peuter: Please change line 38 to "var thisVisu = this;"

ChristianMayer: Done

PR #179, opened 2015-05-06
reenable scope attribute for navbars

peuter:
see: http://knx-user-forum.de/forum/supportforen/cometvisu/829798-darstellungsprobleme-mit-git-version


peuter:
added fix for invisible bottom navbar on initial loading

PR #191, opened 2015-07-23
New plugin for simple links.

stborchert:
Tiny plugin to create simple plugin in CV.

Usage:
```html
<link href=".?forceReload=true" text="Reload" class="button-reload"/>
```


src/plugins/link/structure_plugin.js
```diff
@@ -0,0 +1,41 @@
+/* structure_plugin.js (c) 2012 by Michael Hausl [michael@hausl.com]
```
ChristianMayer: Can you please use the correct informations for you in this line?

PR #196, opened 2015-10-12
Fixed missing var declaration and pragmatic solution regarding OH integration

starwarsfan:
See referenced links on each commit message.


joltcoke:
Label: Bugfix
Milestone: 0.9.0

PR #197, opened 2015-10-12
New plugin to display data from openweathermap.org.

stborchert:
Simple plugin to display data from [openweathermap.org](http://openweathermap.org/).

### Usage
* add the plugin to your configuration (```<plugin name="openweathermap"/>```) 
* embed the plugin in your page configuration:
```xml
<openweathermap detailItems="1" forecastItems="4" lang="de" lat="{LATITUDE}" lon="{LONGITUDE}" refresh="5" appid="{YOUR APP ID}"/>
```

Example output (with some additional styling):
image

### Schema

To pass validation of *check_config.php* you'll need to update the file *visu_config.xsd* as follows.
* add ```<xsd:element name="openweathermap" type="openweathermap" />``` to "AvailablePlugins"
* add the following type definition at the end of the file:
```xml

  <xsd:complexType name="openweathermap">
    <xsd:attribute name="lang" use="optional">
      <xsd:simpleType>
        <xsd:restriction base="xsd:string">
          <xsd:enumeration value="en" />
          <xsd:enumeration value="de" />
          <xsd:enumeration value="fr" />
        </xsd:restriction>
      </xsd:simpleType>
    </xsd:attribute>
    <xsd:attribute name="appid" type="xsd:string" use="required" />
    <xsd:attribute name="detailItems" type="xsd:integer" use="optional" />
    <xsd:attribute name="forecastItems" type="xsd:integer" use="optional" />
    <xsd:attribute name="lat" type="xsd:string" use="optional" />
    <xsd:attribute name="lon" type="xsd:string" use="optional" />
    <xsd:attribute name="refresh" type="xsd:integer" use="optional" />
    <xsd:attribute ref="class" use="optional" />
  </xsd:complexType>
```


src/plugins/openweathermap/owm/jquery.owm.js
```diff
@@ -0,0 +1,269 @@
+/**
+ * Plugin: jquery.OWM
+ *
+ * Version: 0.0.1
+ * (c) Copyright 2015, Stefan Borchert
```
ChristianMayer: Can you please add licence information (should be the JQuery licence or the GPLv3 in this case, I guess)?

ChristianMayer:
Nice plugin!

Are the ".DS_Store" files needed or are they are mistake? If so, please remove them so that I can merge the plugin (after the licence was added :))

stborchert:
* added note to license (GPLv3)
* removed *.DS_Store* files ... (maybe this type of files should be added to a .gitignore file in CometVisu generally?)

MHWallaby:
I think you should add the XSD update as well in this commit, otherwise the usage of the plugin will not work with the editor.

joltcoke:
Label: major-feature
Milestone: 1.0.0

PR #198, opened 2015-10-12
Fixed icon update Script to work with github and updated knx-uf-iconset

starwarsfan:
- Replaced svn call on update script with proper git command and updated internal behaviour to handle git clone instead of svn checkout.

- Updated icons with current state of them


joltcoke:
Label: Enhancement
Milestone: 1.0.0

PR #200, opened 2015-10-20
fix flow animation for svg

cool-runnings:


joltcoke:
Label: Bugfix
Milestone: 0.9.0

ChristianMayer:
I'm not too happy about the chosen solution as it is reactivating a mechanism that was removed.
Currently I'm still reviewing it.

src/lib/templateengine.js
```diff
@@ -402,10 +404,10 @@ function TemplateEngine( undefined ) {
...
   
   this.addAddress = function( address, id ) {
-    if( address in ga_list )
-      ga_list[ address ].push( id );
-    else
+    if ( !(address in ga_list) )
       ga_list[ address ] = [ id ];
+    else if ( $.inArray(id,ga_list[address]) == -1 )
+      ga_list[ address ].push( id );
   };
```
ChristianMayer: What is the reason for this change? Replacing a native JS function by an indirect jQuery function seems like adding only overhead

src/lib/trick-o-matic.js
```diff
@@ -139,4 +141,26 @@ define([ 'jquery' ], function( $ ) {
...
+    $(pipes).each(function(idx, pipe) {
+      var activeValues = $(pipe).data('cometvisu-active');
+      if ( !activeValues ) { return true; }
+      $(activeValues.split(' ')).each(function(idx, ga) {
+        templateEngine.addAddress(ga, 'svg');
+      });
+    });    
+  });
+  jQuery.ajaxSetup({async:true});
+};
```
ChristianMayer: Do I interpret that correctly:
You are reloading the SVG backdrop just to parse it for the GAs? And are using a synchronous call for that?

If that's true I'm sorry to tell that this are two no gos for me:
- the data is already loaded and parsed in the browser and should be accessed through the DOM
- synchronous is considered bad practise in JavaScript and only very special occasions allow it

ChristianMayer:
Please see https://github.com/CometVisu/CometVisu/pull/226 for a different bug fix

cool-runnings:
Maybe this solution wasn't the best idea. I'll have a look for your approach.

PR #202, opened 2015-10-28
fixed wrong comment in css

tuxedo0801:
Just checked out CV and opend as Netbeans HTML5 project, which gave me an error in CSS file. Fixed it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/nbproject/private/
+/nbproject/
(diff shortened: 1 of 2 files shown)


joltcoke:
// isn't a valid CSS comment either. Unfortunately you have to use /* */ style comments.

Could you also please remove the .gitignore file from the pull request?

tuxedo0801:
I'll drop the request and create a new one. Struggling with git commands --> still a git noob :-(

PR #203, opened 2015-10-28
fixed wrong comment in css

tuxedo0801:
Just checked out CV and opend as Netbeans HTML5 project, which gave me an error in CSS file. Fixed it.

Now without gitignore-file ;-)


tuxedo0801:
One commit too much, see http://knx-user-forum.de/forum/supportforen/cometvisu/870953-slider-position-falsch?p=872754#post872754

PR #204, opened 2015-10-28
Fix wrong slider position due to wrong number type. Forcing string-nu…

tuxedo0801:
…mber to decimal/float.


tuxedo0801:
One commit too much, see http://knx-user-forum.de/forum/supportforen/cometvisu/870953-slider-position-falsch?p=872754#post872754

PR #205, opened 2015-10-28
added readonly mode for slider

tuxedo0801:
if readonly, just show the value/slider position, but disable user interaction.

### DIFF
--- a/src/structure/pure/slide.js
+++ b/src/structure/pure/slide.js
@@ -57,6 +57,7 @@ design.basicdesign.addCreator('slide', {
     var max  = parseFloat( $e.attr('max')  || datatype_max || 100 );
     var step = parseFloat( $e.attr('step') || 0.5 );
     var send_on_finish = $e.attr('send_on_finish') || 'false';
+    var readonly = $e.attr('readonly') || 'false';
     var data = templateEngine.widgetDataInsert( path, {
       //???///'events':   $(actor).data( 'events' ),
       'min'            : min,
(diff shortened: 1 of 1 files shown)


joltcoke:
The address item has a mode attribute. Would it be possible to set the slider to readonly if the address mode is "read"?

What happens if readonly is false and mode is read? You still can move the slider but nothing will happen (on the bus)?

joltcoke:
Label: Enhancement
Milestone: 1.0.0

tuxedo0801:
> Would it be possible to set the slider to readonly if the address mode is "read"?

Maybe. My JS skills are not sooo good.

> What happens if readonly is false and mode is read? You still can move the slider but nothing will happen (on the bus)?

with the current PR, if readonly="true", you cannot move the slider at all. It's just showing the value, independant from mode-attribute of address-item.

joltcoke:
var disable = true;

for (var ga in data.address) {
  if(data.address[ga][1] & 2) {
    disable = false;
    break;
  }
}

Should do the trick. You can now use the disable variable to deactivate the widget. sliders without any address will also get disabled.

tuxedo0801:
Could you pls. explain 

```javascript
if(data.address[ga][1] & 2)
```
?

Am I right that the attribute-order or address-item is relevant?

joltcoke:
Sure!

data.address contains all GA's associated with the widget. Since it is an Javascript Object you can not access it with numeric indices. You either have to use a for ... in loop or know the index beforehand.

Index is the address name as supplied in the config XML. Each address will have three fields, indexed numerically:

0: transform
1: disable/read/write/readwrite mode (Bit 0: read, Bit 1: write)
2: variant

Attribute order in the config file is not relevant, it will always mapped according to the above rules. The code i've posted above will loop over all addresses configured for the widget. If there is at least one with write access, disabled will be false. If there is no address at all or all of them are configured read only, disable will be true.

tuxedo0801:
Thanks for explanation. Tested, Works. Will update the PR.

tuxedo0801:
implementation updated accordingly.

PR #206, opened 2015-10-28
Fix wrong slider position due to wrong number type. Forcing string-nu…

tuxedo0801:
…mber to decimal/float.


joltcoke:
Looks good from my end! (Bugfix / 0.9.0)

PR #208, opened 2015-10-28
Multi GA support for info widget

joltcoke:
As discussed here: http://knx-user-forum.de/forum/supportforen/cometvisu/870759-info-werte-zusammensetzen?p=872524#post872524


joltcoke:
Label: Enhancement
Milestone: 1.0.0

joltcoke:
Questions:

1. I've added just the min version of the sprintf library. Should i also add the non min version?
2. Do i need to modify the appcache file and list the additional files? Which ones?

joltcoke:
Replaced by #210

PR #209, opened 2015-10-28
added missing session to write-command

tuxedo0801:
write command was missing the session variable.

Even if session is not well supported by current backends, it should be possible for any backend to renew it's internal session timer/timeout when processing a write-command based in provided session.


joltcoke:
Seems to be in line with the protocol documentation => Bugfix.

Label: Bugfix
Milestone: 0.9.0

PR #224, opened 2015-11-29
Comma cleanup

ChristianMayer:
Some browsers / JavaScript engines don't like the trailing commas.


src/structure/pure/slide.js
```diff
@@ -182,6 +182,15 @@ design.basicdesign.addCreator('slide', {
...
   },
+  sliderVisible:function(event,page_id)
+  {
+    $('.ui-slider-handle.untransformed', '#'+page_id).each(function(i) {
+      $(this).removeClass('untransformed');
+      var actor = $(this).parent();
+      var val = actor.slider("value");
+      transformSlider(val,this);
+    });
+  }
```
peuter: Whats this about? At least it doesn't seem to belong to the comma cleanup PR, right?

ChristianMayer: Argh! This block of code is haunting me :(
No, it doesn't belong here

PR #226, opened 2015-11-29
Fix flow animation for svg backdrop (trick-o-matic)

ChristianMayer:
Transfer event handling to the new structure that was introduced with the textify.

(Note: this is an alternative fix to the pull request https://github.com/CometVisu/CometVisu/pull/200 )


cool-runnings:
This works so far. The only problem is, that when I open the page the animation is not working altough the GA is 1. Only when the backend receives an update on the GA, the animation starts. That's why I had the trickomatic_initGAs function added.

PR #227, opened 2015-12-17
Initial contribution of the new universal client

peuter:
- one client to rule them all
- including autoconfiguring by backend


tuxedo0801:
From my point of view: looks almost stable. No negative behavior found. 
Runs since weeks without bigger issues. 
The only thing i see from time to time: auto-reconnect is not working as expected. As this issue is very sporadic, it's hard to debug. No reproductions scheme known yet.

ChristianMayer:
Do I understand correctly that `universal-client.js` is basically the existing `cometvisu-client.js` just with an additional transport protocol?
If yes: I guess it's better to implement that functionality in the existing file

And when I understand it correctly you shouldn't name the new transport layer "openhab*" as that's what it's beeing used for and not what it's doing... :)

tuxedo0801:
> Do I understand correctly that universal-client.js is basically the existing cometvisu-client.js just with an additional transport protocol?

Basically, from an abstract point of view, yes. But technically, is a clean, fresh and well structured implementation that does not really share a lot of code with the old implementation. 

> If yes: I guess it's better to implement that functionality in the existing file
To have it in the same file would mean: throw away all the content of cometvisu-client.js and use the new implementation. So it's just a matter of file-naming.

> And when I understand it correctly you shouldn't name the new transport layer "openhab*" as that's what it's beeing used for and not what it's doing... :)

The new transport layer is Server-Sent-Events, not openhab. "openhab" is the preconfigured backend-name, not the transport. 

@peuter correct me if I'm wrong ...

br,
  Alex

peuter:
@tuxedo0801 Couldn't have said it better ;-) Only one addition, despite the new transport layer there are also multiple possibilities added to configure the client, by hardcoding or sent by backend.

src/lib/templateengine.js
```diff
@@ -274,33 +276,20 @@ function TemplateEngine( undefined ) {
   
   this.enableAddressQueue = $.getUrlVar('enableQueue') ? true : false;
   
-  this.backend = 'cgi-bin'; // default path to backend
+  this.backend = 'default';
+  this.backendUrl = null;
```
ChristianMayer: `null` is quite uncommon in JS. `undefined` is the usual way to say it's not set. (This does not mean `null` is wrong. But uncommon might come unexpected)

src/lib/universal-client.js
```diff
@@ -0,0 +1,579 @@
...
+  
+/**
+ * Class that handles the communicaton of the client * 
+ * @class CometVisu
+ * @constructor foo
+ * @param backend
+ */
+function CometVisu( backend, initPath )
+{
+  this.initPath = initPath ? initPath : null;
```
ChristianMayer: See above about `null`

src/lib/universal-client.js
```diff
@@ -0,0 +1,579 @@
...
+  		resendHeaders : {
+  		  'X-Atmosphere-tracking-id' : null
+  		},
+  		// fixed headers that are send everytime
+  		headers : {
+  		  'X-Atmosphere-Transport' : 'long-polling'
+  		},
+  		hooks : {
+  		  onClose : function() {
+  		    // send an close request to the openHAB server
```
ChristianMayer: Please replace Tabs by spaces - otherwise the code alignment might break as demonstrated on this line above

src/lib/universal-client.js
```diff
@@ -0,0 +1,579 @@
...
+  	      this.headers["X-Atmosphere-Transport"]="close";
+  	      $.ajax( {url:this.config.baseURL + this.config.resources.read, dataType: 'json', context:this, beforeSend:this.beforeSend } );
+  	      if (oldValue!=undefined) {
+  	        this.headers["X-Atmosphere-Transport"]=oldValue;
+  	      }
+  	      else {
+  	        delete this.headers["X-Atmosphere-Transport"];
+  	      }
+  		  }
+  		}
```
ChristianMayer: -Tab +Space

src/lib/templateengine.js
peuter: In this case I want to intialize a variable with no value set, what is exactly what 'null' is used for (also in Javascript).
So this.backendUrl === null is an expected (because set by user) behaviour, whereas this.backendUrl === undefined would mean something went terribly wrong. To distinguish these cases the initialization of variables with 'null' is the way to go from my point of view.

ChristianMayer: As I said, it's not wrong to use `null`. See e.g. http://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript for a deeper discussion. The accepted answer has IMHO a good conclusion:
- `undefined` means a variable has been declared but has not yet been assigned a value
- `null` is an assignment value. It can be assigned to a variable as a representation of no value

I'm fine with `null` when that's the logical value here. When you want to say "default" for being not set, then in my opinion `undefined` would be the logical choice.

peuter: Thats the link I read before posting ;-) It sais there:

```
You might consider undefined to represent a system-level, unexpected, or error-like absence of value and null to represent program-level, normal, or expected absence of value.
```
Thats why I prefered to go with 'null', but this is rather an academical discussion and because it fits the cometvisu coding style better I will remove the null's

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+                this.xhr = $.ajax({
+                  url : thisCometVisu.getResourcePath("read"),
+                  dataType : 'json',
+                  context : this,
+                  data : thisCometVisu.buildRequest() + '&i=' + this.lastIndex,
+                  success : this.handleRead,
+                  error : this.handleError,
+                  beforeSend : this.beforeSend
+                });
+                this.watchdog.ping();
```
ChristianMayer: Running the code (eibd backend, no change to my config, just a local pull of this request) I get a error in this line:
`Uncaught TypeError: this.watchdog.ping is not a function`

(Why? I don't know yet...)

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+          /**
+           * ts is a quirk to fix wrong caching on some Android-tablets/Webkit;
+           * could maybe selective based on UserAgent but isn't that costly on writes
+           */
+          var ts = new Date().getTime();
+          $.ajax({
+            url : this.getResourcePath("write"),
+            dataType : 'json',
+            context : this,
+            data : 's=' + session + '&a=' + address + '&v=' + value + '&ts='
```
ChristianMayer: Failure: `Uncaught ReferenceError: session is not defined`

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+ * 
+ * @module CometVisu Client
+ * @title CometVisu Client
+ * @reqires jQuery
+ */
+define(
+    [],
+    function() {
+      "use strict";
+
```
ChristianMayer: Geschmackssache: durch diese Anordnung haben wir bereits 6 Zeichen Einzug ohne das wir dadurch Information / Struktur haben.
Ob man nach dem globalen `define` mit einem Einzug weiter macht oder nicht ist sicher streitbar, 3x Einzug halte ich persönlich für übertrieben.

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+          }
+          // add trailing slash to baseURL if not set
+          if (this.config.baseURL && !this.config.baseURL.endsWith("/")) {
+            this.config.baseURL += "/";
+          }
+        };
+        this.checkSettings();
+
+        /**
+         * Subscribe to the addresses in the parameter The second parameter
```
ChristianMayer: '.' missing?

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+        if (backend && backend !== 'default') {
+          if ($.isPlainObject(backend)) {
+            // override default settings
+            $.extend(this.config, backend);
+          } else if (this.backends[backend]) {
+            // merge backend settings into this.config
+            $.extend(this.config, this.backends[backend]);
+          }
+        }
+
```
ChristianMayer: Quite often in the code below a structure like `this.transport[this.config.transport]. ...` is used.

To help readability and performance (a very little) it could be a good idea to introduce a local variable here that contains `this.transport[this.config.transport]` and is used further down. (Due to the refcounting nature of JS the memory overhead is minimal)

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+       */
+      function CometVisu(backend, initPath) {
+        this.initPath = initPath;
+
+        // used for backwards compability
+        this.aliases = {
+          'cgi-bin' : 'default',
+          'oh' : 'openhab',
+          'oh2' : 'openhab2'
+        };
```
ChristianMayer: Geschmacksfrage: Ich bevorzuge es, wenn in einem Objekt/Hash die Werte für die Keys linksbündig sind, in diesem Fall wäre dass dann:

    this.aliases = {
      'cgi-bin' : 'default',
      'oh'      : 'openhab',
      'oh2'     : 'openhab2'
    };

Bei diesem kleinen Objekt ist das nicht wirklich relevant, aber weiter unten (z.B. wo die Ajax-Requests zusammen gebaut werden) schaut's für mir unordentlich und schwer zu lesen aus

ChristianMayer:
For a general discussion and comment please see http://knx-user-forum.de/forum/supportforen/cometvisu/874578-universalclient?p=889738#post889738

src/lib/universal-client.js
```diff
@@ -0,0 +1,638 @@
...
+                  // delete last;
+                  last = new Date();
+                  if (!this.doRestart) {
+                    // delete hardLast;
+                    hardLast = last;
+                  }
+                }
+              };
+            },
+
```
ChristianMayer: To get the watchdog to run (see error on line 371), I had to change this block to:

    watchdog : (function() {
      var last = new Date();
      var hardLast = last;
      var aliveCheckFunction = function() {
        var now = new Date();
        if (now - last < thisCometVisu.transport['long-polling'].maxConnectionAge * 1000)
          return;
        if (now - hardLast > thisCometVisu.transport['long-polling'].maxDataAge * 1000)
          thislastIndex = -1; // reload all data
        thisCometVisu.transport['long-polling'].restart();
        last = now;
      };
      setInterval(aliveCheckFunction, 5 * 1000);
      return {
        ping : function() {
          // delete last;
          last = new Date();
          if (!thisCometVisu.transport['long-polling'].doRestart) {
            // delete hardLast;
            hardLast = last;
          }
        }
      };
    })(),

but you see it's quite ugly as by using the `{ key1: value, key2: value }` notation in this case doesn't allow direct access to the other values, especially not during setup.

I guess it's better to change that to a real JS Object with constructor and local variables as there's no need to access them from the outside.

(Note: the `cometvisu-client.js` was written by me on a time where my understanding of JS wasn't as deep as it is today. So there are a few places that should be a bit different when modern JS is used. There's no need to copy the bad parts :) )

peuter:
Further development happens in new feature branch 'universal-client' in the CometVisu repository

PR #230, opened 2015-12-30
Rsslog improvements

ChristianMayer:
rsslog database:
- Use database to keep track of schema version
- Add mapping column when version is too old

rsslog:
- translate demo config to english
- make itemack configurable
- implement limit
- implement mapping so that icons can be used
- allow address elements - use this to notify about data change so that the data can be reloded


peuter:
As this is a rather complex PR with lots of changes I would prefer to test is before merging. But with my current openHAB environment I cannot test the php-related parts of the CometVisu.

So my question is how we should proceed here? I can merge it with a manual code review, without test or someone else has to do the test/merge here.

ChristianMayer:
I don't know how many people can do the review.

So I suggest:
- do a best effort review
- stuff in a plugin have less severe review requirements as it's optional stuff - in comparison to normal widgets. (The cometvisu-client has the most severe requirements as it's fundamental)
- in doubt about something: ask and we'll discuss

PR #232, opened 2016-01-02
Fix demo with metal

ChristianMayer:
Opening the (normal) demo config with the metal design in Chrome didn't
show the main screen as it was too wide and the browser placed it to the
bottom.

This patch fixes the issue by forcing a recalculation.

A real fix would be to follow though the page creation, especially when
the navbars are active, and handle the sized. This is kept for future
development...

### DIFF
--- a/src/lib/templateengine.js
+++ b/src/lib/templateengine.js
@@ -768,7 +768,7 @@ function TemplateEngine( undefined ) {
     // the calculation has to be done again, even if the page hasn´t changed (e.g. switching between portrait and landscape mode on a mobile can cause that)
     var bodyWidth = $('body').width();
     var mobileUseChanged = (lastBodyWidth<thisTemplateEngine.maxMobileScreenWidth)!=(bodyWidth<thisTemplateEngine.maxMobileScreenWidth);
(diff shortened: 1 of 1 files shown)


src/lib/templateengine.js
```diff
@@ -768,7 +768,7 @@ function TemplateEngine( undefined ) {
     // the calculation has to be done again, even if the page hasn´t changed (e.g. switching between portrait and landscape mode on a mobile can cause that)
     var bodyWidth = $('body').width();
     var mobileUseChanged = (lastBodyWidth<thisTemplateEngine.maxMobileScreenWidth)!=(bodyWidth<thisTemplateEngine.maxMobileScreenWidth);
-    if (thisTemplateEngine.currentPageUnavailableWidth<0 || mobileUseChanged) {
+    if (thisTemplateEngine.currentPageUnavailableWidth<0 || mobileUseChanged || true) {
```
peuter: Wouldn't it be easier if you just comment out the if statement as it always returns true.

ChristianMayer: Yes, but as I don't know exactly why we are checking this only conditionally (performance?) I wanted to keep it as a reminder...

PR #238, opened 2016-01-16
Manager default fix

ChristianMayer:
See also http://knx-user-forum.de/forum/supportforen/cometvisu/870752-fehler-im-visu_config-xml-beseitigen/


peuter:
I guess the new design shouldn't be part of this PR, right?

ChristianMayer:
You are right. I'll create a new PR

PR #241, opened 2016-01-23
touchfix + new maxMobileScreenWidth

alltime84:
- Increased maxMobileScreenWidth to use mobile.css even if smartphone is turned to landcape
- added threshold for touchmove to not cancel clicks on minor moves (touch click did not react on every click)

### DIFF
--- a/src/lib/templateengine.js
+++ b/src/lib/templateengine.js
@@ -499,26 +499,26 @@ function TemplateEngine( undefined ) {
         downtime:        0,
         alreadyCanceled: false
       };
-			
-		var touchStartX = null;
-		var touchStartY = null;
-		
+
+    var touchStartX = null;
+    var touchStartY = null;
+
     window.addEventListener( isTouchDevice ? 'touchstart' : 'mousedown', function( event ){
       var 
         element = event.target,
         // search if a widget was hit
         widgetActor = getWidgetActor( event.target ),
         bindWidget  = widgetActor.widget ? thisTemplateEngine.widgetDataGet( widgetActor.widget.id ).bind_click_to_widget : false;
-				
-			var touchobj;
-			
-			if (isTouchDevice){
-				touchobj = event.changedTouches[0];
-				
-				touchStartX = parseInt(touchobj.clientX);
-				touchStartY = parseInt(touchobj.clientY);
-			}
-			
+      
+      var touchobj;
+      
+      if (isTouchDevice){
+        touchobj = event.changedTouches[0];
+        
+        touchStartX = parseInt(touchobj.clientX);
+        touchStartY = parseInt(touchobj.clientY);
+      }
+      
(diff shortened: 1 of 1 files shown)


ChristianMayer:
Can you please replace the Tab characters by Space (cf. https://github.com/CometVisu/CometVisu/wiki/Coding-style)

alltime84:
ahhhh.... OK, I hope it fits to the coding style now.

src/lib/templateengine.js
```diff
@@ -687,7 +703,7 @@ function TemplateEngine( undefined ) {
       if (!ret && m['defaultValue']) {
         ret = mapValue(m['defaultValue']);
       }
-      if (ret) {
+      if( ret !== undefined ) {
```
ChristianMayer: As this is included here and not on the original version it seems that this patch is against a old version. (GitHub tells me that it is CometVisu/CometVisu develop but that's not correct when you refer to https://github.com/CometVisu/CometVisu/blame/develop/src/lib/templateengine.js )
=> Can you rebase that request on the latest CometVisu/CometVisu develop?

src/lib/templateengine.js
```diff
@@ -498,7 +499,10 @@ function TemplateEngine( undefined ) {
         downtime:        0,
         alreadyCanceled: false
       };
-    
+
+    var touchStartX = null;
```
ChristianMayer: please don't use new `var` for each line. Just reuse the `var` above from line 488 ff.

src/lib/templateengine.js
```diff
@@ -216,7 +216,8 @@ function TemplateEngine( undefined ) {
   this.bindClickToWidget = false;
     
   // threshold where the mobile.css is loaded
-  this.maxMobileScreenWidth = 480;
+	// this.maxMobileScreenWidth = 480; // modified for Menu Plugin!
+  this.maxMobileScreenWidth = 780;
```
ChristianMayer: Do you really need that? Woudn't a config file with the attribute `max_mobile_screen_width` set in the `<pages>` element be sufficient?

When you do need it: please don't keep the old version in the comment in the line above (219)

ChristianMayer:
I've just tested it and the patch functionality seems fine to merge. 
Please check my line comments and either change the code or tell me why you think it's fine there.
Then I can merge the code.

ChristianMayer:
@alltime84 as my pull request #242 was merged earlier and thus created a conflict for this patch, I suggest that I will resolve the collision once you've  answered my points. (Of course you are free to resolve the collision as well :))

src/lib/templateengine.js
alltime84: Fine for me. No need to be hardcoded. I just did not know it exists.

src/lib/templateengine.js
alltime84: Ok, feel free to change it.

alltime84:
Please go ahead.

ChristianMayer:
It's merged now

PR #259, opened 2016-02-07
remove android specific coloring method

alltime84:
Android-fix not required anymore for current Android versions.


ChristianMayer:
You must keep that code for older Android versions.

According to https://code.google.com/p/android/issues/detail?id=17565 it was fixed with Android 4.4.
According to http://developer.android.com/about/dashboards/index.html of today about 1/3 of the currently used devices would be broken by this change. (And assuming our audience with wall mounted devices it can be assumed that they are using an over-proportional share of older devices)

=> Please leave this code active and switch it to version detection.

alltime84:
ok, switched to version detection like suggested.

alltime84:
And sorry, I mixed up this pull request with another change (fix ro enable scrollable navbarLeft). I am beginner using git...

ChristianMayer:
No worries, everybody had to start once - and with Git I'm also still quite a beginner.

But please get the scrollbar out of this patch to keep it clean.

(Brute force beginner way could be to throw this branch away and start new ;))

PR #261, opened 2016-02-07
enable scrollable navbarLeft on mobile devices

alltime84:
this is a fix to enable a scrollable navbar left on mobile devices. I am sure pageparthandler.js is not the best file to place the function call, so please help to find the correct file/position.


alltime84:
Please ignore this pull-request. It is not working without css adjustments.
I will create a new pull-request with an mobilemenu-plugin, which (hopefully) enables scrolling and an android like sidemenu.

alltime84:
I created a new plugin called mobilemenu. This plugin enables android like menu you can show and hide. Furthermore it enables a scrollable navbarLeft.

src/designs/pure/mobile.css
```diff
@@ -9,3 +9,10 @@
 .page {
     overflow: visible;
 }
+
+#navbarLeft .navbar{
+  background-color: #000000;
+  -webkit-box-shadow: 1px 0px 16px gray;
+  -moz-box-shadow: 1px 0px 16px gray;
+  box-shadow: 1px 0px 16px gray;
+}
```
ChristianMayer: Does this patch need this change to the pure/mobile.css design? I guess it doesn't belong to this patch, does it?

src/plugins/mobilemenu/mobilemenu.css
```diff
@@ -0,0 +1,22 @@
+.mobilemenu{
+  position: absolute !important;
```
ChristianMayer: The use of `!important` should be very limitied as it is already an escalation mechanism.
Probably it's better to make the CSS rule more specific to make it win over the other rules. This would allow the use of `!important` for those cases where it really is important, like in a custom.css config.

src/designs/pure/mobile.css
alltime84: No, I just added it to bettet show the overlapping navbar. I will remove it before a merge.

alltime84:
I removed the css from mobile.css because it was only used for demo reasons. Furthermore I was able to get rid of some "!important" tags.

Issue #262, opened 2016-02-07
sprintf troubles with multiple addresses


ChristianMayer:
As described in http://knx-user-forum.de/forum/supportforen/cometvisu/907905-infotrigger-zeigt-wert-nicht-an-wenn-mehr-als-eine-ga-definiert the sprintf has problems when multiple addresses are used.

In the example given it's a problem that the first address is a read only address and thus the parameter to the sprintf will have a '<UNDEF>' for that - which breaks as the format string is expecting a number there.

ChristianMayer:
A different example that currently fails:
```
      <info format="%.2f°C">
        <label>Alle</label>
        <address transform="DPT:9.001" mode="read">12/7/91</address>
        <address transform="DPT:9.001" mode="read">12/7/92</address>
        <address transform="DPT:9.001" mode="read">12/7/93</address>
      </info>
      <slide min="10" max="30">
        <label>actual</label>
        <address transform="DPT:9.001" mode="readwrite">12/7/91</address>
      </slide>
      <slide min="10" max="30">
        <label>setpoint</label>
        <address transform="DPT:9.001" mode="readwrite">12/7/92</address>
      </slide>
      <slide min="0" max="100">
        <label>control</label>
        <address transform="DPT:9.001" mode="readwrite">12/7/93</address>
      </slide>

```

PR #263, opened 2016-02-09
Ignore write only addresses when formatting strings

joltcoke:
Fix for #262


ChristianMayer:
As described in http://knx-user-forum.de/forum/supportforen/cometvisu/907905-infotrigger-zeigt-wert-nicht-an-wenn-mehr-als-eine-ga-definiert?p=911803#post911803 this patch doesn't allow multiple GAs be used for the same value ("object" in KNX speech).

ChristianMayer:
https://github.com/CometVisu/CometVisu/pull/281 solved that issue

PR #267, opened 2016-02-14
Page change callbacks

ChristianMayer:
This also fixes that the RRDs were constantly updated after the first
view even when the widget was not visible (i.e. not on the current page)
anymore.

It also helps to reduce traffic when multiple diagrams are using the
same data by caching it.

This should fix https://github.com/CometVisu/CometVisu/issues/244
This should create infrastructure to be able to fix https://github.com/CometVisu/CometVisu/issues/243 - but this issue is not fixed by this pull request yet

### DIFF
--- a/src/lib/templateengine.js
+++ b/src/lib/templateengine.js
@@ -288,6 +288,8 @@ function TemplateEngine( undefined ) {
    */
   var pathRegEx = /^id(_[0-9]+)+$/;
 
+  var currentPath = '';
+  this.callbacks = {}; // Hash of functions to call during page change
   this.main_scroll;
   this.old_scroll = '';
   this.visu;
@@ -1311,6 +1313,9 @@ function TemplateEngine( undefined ) {
       $('.pageActive', '#pages').removeClass('pageActive');
       thisTemplateEngine.currentPage.addClass('pageActive activePage');// show new page
       $('#pages').css('left', 0 );
+      thisTemplateEngine.callbacks[currentPath].afterPageChange.forEach( function( callback ){
+        callback( currentPath );
+      });
     });
     if (thisTemplateEngine.scrollSpeed != undefined) {
       thisTemplateEngine.main_scroll.getConf().speed = thisTemplateEngine.scrollSpeed;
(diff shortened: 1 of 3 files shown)


src/lib/templateengine.js
```diff
@@ -1371,6 +1376,14 @@ function TemplateEngine( undefined ) {
 
   this.create_pages = function(page, path, flavour, type) {
     var creator = thisTemplateEngine.design.getCreator(page.nodeName);
+    
+    thisTemplateEngine.callbacks[ path + '_' ] = {
+      exitingPageChange: [],// called when the current page is left
+      beforePageChange: [], // called as soon as a page change is known
+      duringPageChange: [], // called when the page is theoretical visible, i.e. "display:none" is removed - CSS calculations shoud work now
+      afterPageChange: []   // called together with the global event when the transition is finished
```
peuter: Do we really need before, during and afterPageChange? What would happen if strip them down to only
 enteringPage, leavingPage (and maybe beforeEnteringPage) (IMHO those names would fit better)?

ChristianMayer: The idea is that communication to get the necessary content can already be started by `beforePageChange`. So this is called as soon as the change is known.  
The `duringPageChange` comes shortly afterwards. Here the `display:none` is already removed so the CSS measurements should work now. This also means that a reflow has happend. This could take some time - not much, but during that time the server could already process the response.  
The `afterPageChange` happens once the page is finally there as it should be. This means that the page transision / animation time has passed.

In the past only the last event was available - and that "only" via a "slow" global event. The `afterPageChange` uses a direct callback, i.e. is quite lightweight.  
As it comes quite late it's definitely a different time point than the other callbacks.

`beforePageChange` and `duringPageChange` are really close in time. So they could be unified to the `duringPageChange` (a unification to `beforePageChange` can not work as CSS measurements won't work at that time).  
But I recon the little speed advantage of kicking of the communication before and parallel to the reflow is worth the ~5 lines of additional code.

peuter: Ok thanks for clearing that up.

cool-runnings:
After this commit diagrams are no longer loaded if they are within a group.

ChristianMayer:
OK, I'll investigate

ChristianMayer:
Bug confirmed

ChristianMayer:
Pull-Request for the fix: https://github.com/CometVisu/CometVisu/pull/270

PR #269, opened 2016-02-19
Add display of future events to rsslog

cool-runnings:
Added the parameter future to show upcoming events (e.g. garbage collections) in the rsslog.
Fixed the following:
- Separator before first entry
- rsslog_prevday was never set back to false

image


src/plugins/rsslog/structure_plugin.js
```diff
@@ -54,7 +54,8 @@ VisuDesign_Custom.prototype.addCreator("rsslog", {
       limit:      $el.attr("limit") ? +$el.attr("limit") : 0,
       timeformat: $el.attr("timeformat"),
       itemoffset: 0,
-      itemack:    $el.attr("itemack") || "modify" // allowed: modify, display, disable
+      itemack:    $el.attr("itemack") || "modify", // allowed: modify, display, disable
+	  future:     $el.attr("future"),
```
ChristianMayer: Please use spaces instead of tab

src/plugins/rsslog/structure_plugin.js
```diff
@@ -123,6 +124,14 @@ function refreshRSSlog( data, isBig ) {
         src += '?limit=' + limit;
       }
     }
+	var future = data.future;
```
ChristianMayer: Please use spaces instead of tab

src/plugins/rsslog/structure_plugin.js
```diff
@@ -260,13 +271,21 @@ function refreshRSSlog( data, isBig ) {
...
                 templateEngine.design.defaultValue2DOM( mappedValue, function(e){ $span.append(e); } );
               }
-              if (separatoradd) { 
+              if (separatoradd & idx !== 0) { 
                 $row.addClass('rsslog_separator');
                 separatorprevday = true; 
               }
+			  else {
+                separatorprevday = false;
+			  }
```
ChristianMayer: Lines 278 + 280: Please use spaces instead of tab

src/plugins/rsslog/rsslog.php
```diff
@@ -319,6 +321,11 @@ function retrieve( $db, $filter, $state )
   
   if (isset($state))
     $q .= " AND state=" . $state . " ";
+
+  if (isset($future))
+    $q .= " AND ((t  <= datetime('now','+" . $future . " hour') )) ";
```
ChristianMayer: Problem for SQL injection attack?
=> Please ensure that `$future` is a number

PR #270, opened 2016-02-19
Fix widgets with callback in groups

ChristianMayer:
See https://github.com/CometVisu/CometVisu/pull/267

### DIFF
--- a/src/lib/templateengine.js
+++ b/src/lib/templateengine.js
@@ -1419,7 +1419,7 @@ function TemplateEngine( undefined ) {
       parent = element.parentNode,
       parentpath = widgetpath.replace( /[0-9]*$/, '' );
     
-    while( parent.nodeName === 'group' )
+    while( parent && parent.nodeName !== 'page' )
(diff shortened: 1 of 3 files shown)


src/lib/templateengine.js
```diff
@@ -1407,6 +1407,26 @@ function TemplateEngine( undefined ) {
...
+   * @param widgetpath The path / ID of the widget
+   * @return The path of the parent
+   */
+  this.getPageIdForWidgetId = function( element, widgetpath )
+  {
+    var
+      parent = element.parentNode,
+      parentpath = widgetpath.replace( /[0-9]*$/, '' );
+    
+    while( parent.nodeName === 'group' )
```
peuter: As there are other elements that can contain widget-elements like navbars, infoaction-plugin I guess.`while (parent && parent.nodeName !== 'page')` would be better, right?

ChristianMayer: Good point

PR #272, opened 2016-02-19
Fix tabs

ChristianMayer:
Bring the code in a shape that the new Travis CI code style check (which looks for lines starting with a tabulator) is happy


peuter:
I guess you forget to check in the moved files in the dep directories as I see only deletions, right?

ChristianMayer:
You were right. Just pushed them.

PR #289, opened 2016-03-13
Add retry link to error message

ChristianMayer:
Usecase: when the visu is shown in kiosk mode and the server can't deliver
the config (e.g. after power failure when visu and server are restarting)
there is currently only the error message shown. There is no option
to reload the page when the config gets available again.
As there's no other interaction possibility the visu currently must be
restarted.

This fix helps by adding a possibility to reload the page from the
error message itself.

### DIFF
--- a/src/lib/templateengine.js
+++ b/src/lib/templateengine.js
@@ -151,6 +151,7 @@ require([
         else
           message += '.';
     }
+    message += '<br/><br/><a href="">Retry</a>';
     $('#loading').html(message);
   };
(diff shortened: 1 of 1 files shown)


ChristianMayer:
@peuter This is somewhere between bug fix and enhancement.
=> Please judge yourself if you want to merge it into the 0.9.1, i.e. if it violates the feature freeze or not

peuter:
As this is a very small fix/enhancement with very little impact I think it is ok to merge into 0.9.1

PR #285, opened 2016-03-07
fix styling of widgetinfos in pagejumps by adding the infoaction clas…

peuter:
…s to the pagejump widget

see: https://knx-user-forum.de/forum/supportforen/cometvisu/824414-vorstellung-infoaction-plugin?p=923187#post923187

PR #290, opened 2016-03-14
fixed styling regression of widgetinfo elements in navbar-pagejumps i…

peuter:
…ntroduced with:

https://github.com/CometVisu/CometVisu/pull/285


ChristianMayer:
Ah, I've just merged that into the 0.9.1-branch past the release.

=> @peuter Please do an additional pull request against develop, so that this patch doesn't get lost.

PR #295, opened 2016-03-20
Convert IconSet to webfont

### LINKED ISSUE #251: Use icons as webfonts
It is possible to create a webfont from the used svg icons. The main advantage of this feature would be that all icons can be scaled and colored easily (just like a font), it would decrease the size of the cometvisu package as the png versions of the icons wouldn't be needed anymore.

The webfont generation is quite easy done via grunt as you can see here https://github.com/peuter/knx-uf-iconset (just a quick proof-of-conecpt, needs some finetuning).

What needs to be done is configuring the build process for the special needs of the cometvisu 
and change the iconhandler to use the icons from the webfont.

peuter:
including grunt task to generate the font

closes #251


ChristianMayer:
Running the Grunt taks I get 
Running "exec:make_woff" (exec) task
```
>> Exception in thread "main"
>> java.lang.UnsupportedClassVersionError: com/google/typography/font/tools/sfnttool/SfntTool : Unsupported major.minor version 52.0
```

ChristianMayer:
And also the icon size changed in comparison to the original ones.

ChristianMayer:
I just reverted that pull as it's result is IMHO not stable enough at the moment.
(Probably it should also mature in a separate branch?)

peuter:
The first problem seems to be related to the java version you are using (which one is it?).

The size problem is a bigger showstopper. I did some further investigation on this. It seems as if the svg icons do not make full use of their defined viewPort, which means that there is lots ob padding around the "real" icon. Therefore this free space makes the icons smaller. There is not much you can do from the outside, IMHO this problem must be solved in the SVG sources, maybe @manoficons can help here?

peuter:
I just pushed the branch to the cometvisu repo, including a workaround for the size problem. I manually modify the viewPort in the SVG files and add a transform to them, which removes the 69.5px padding around the icons. But I am not really sure if this is the right solution, because it is really an ugly workaround, which might introduce new problems.

PR #296, opened 2016-03-20
Feature/unit tests

peuter:
introduces unit-testing (grunt karma:travis) and end-to-end tests (grunt e2e) to the CometVisu.

Some basic tests are implemented to show how its done as a basis for further tests. Tests can be found in the test subdirectory (*.spec.js files). They are split into karma (unit-tests) and protractor (end-to-end tests).

PR #304, opened 2016-04-02
Introduce automated testing

peuter:
unit-testing (grunt karma:travis) and end-to-end tests (grunt e2e) to the CometVisu.

Some basic tests are implemented to show how its done as a basis for further tests. Tests can be found in the test subdirectory (*.spec.js files). They are split into karma (unit-tests) and protractor (end-to-end tests).

follow up to #296


Gruntfile.js
```diff
@@ -125,7 +168,7 @@ module.exports = function(grunt) {
           //'icon/iconconfig.js',
           'lib/templateengine.js',
           'designs/**/*.*',
-          'plugins/**/*.{js,css,png,jpf,ttf,svg}'
+          'plugins/**/*.{js,css,png,jpf,ttf,svg,map}'
```
ChristianMayer: Should the map files really be included in the appcache?

peuter: Why shouldn`t they?

peuter: Just saw the jpf entry in that line, shouldn't that be jpg instead?

PR #305, opened 2016-04-03
always listen to mouse + touch events

### LINKED ISSUE #302: Mouse doesn't work if touch is detected
on my Windows Tablet the mouse doesnt have an effect if a tocuscreen is detected.
workarround by disableing the touchscreen via Dev-Mngr.

Windows 10
Chrome 49.0.2623.110

Auf meinem Windows Tab kann ich sobald ein Touchscreen erkannt wird, die Maus in cometvisu nicht mehr verwenden.
nach deaktivieren im Gerätemanager Funtioniert es.

peuter:
The main problem is that if the CometVisu detects a touch-device is listens to touch* events. If you are using the mouse on the touch-devive (e.g. a windows 10 tablet) those mouse-events are not obeyed. Additionaly there are browsers which are detected as touch-enabled event when there is no touch device (e.g. Midori browser on Linux), which also do not work.

This fix has been tested on Chrome (Linux no touch, Win10-Touchscreen), Android smartphone and Midori (Linux, no touch). 

fixes #302


ChristianMayer:
Will the merge conflict be resolved or shall I review it as it is before it get's resolved?

peuter:
Sorry I didn't realize that there were merge conflicts. They have been solved now.

ChristianMayer:
Test with Chrome 49 on Galaxy Note 10.1 (2014 Edition) shows that this patch is breaking the button handling:

- Open Demo config without the patch:
- go to the first trigger
- press it - it goes immediately down
- keep the press and move it out of the button - the button goes immediately up again

With the patch:
- press the trigger - it take about 1 second until it's appearing pressed (for information: just hitting and releasing it does send the trigger, i.e. it's just an optical problem here)
- sliding the finger out of the pressed button will cause the button to stick pressed
- this can only be resolved by tapping that button again
- I get this error message on the console:

> templateengine.js?_=1460841338210:508 Uncaught TypeError: Cannot read property '0' of undefined
> (anonymous function)	@	templateengine.js?_=1460841338210:508
> dispatcher.dispatchEvent	@	pep.js?_=1460841338210:502
> dispatcher.fireEvent	@	pep.js?_=1460841338210:423
> dispatcher.move	@	pep.js?_=1460841338210:321
> mouseEvents.mousedown	@	pep.js?_=1460841338210:847
> dispatcher.eventHandler	@	pep.js?_=1460841338210:372

Even worse was the test on the iPad: the normal buttons (trigger, switch, ...) are not reacting to the touching at all. Neither animation nor sending of values.

peuter:
I did a complete rework of this fix. The formerly used polyfill has been deleted as it did some timeout stuff to detect some touch gestures, which we don't need and which I wasn't able to change there.

So this try is more related to the original implementation, despite beeing refatored the main change is that we are listening both the moust+the touch events everytime. Because there are devices which can use both. The main problem is that we have to make sure, that not both events are handled for one click. Somehow using event.preventDefault in an touchend event prevented all mouseevents from beeing fired after the touch events (which is waht we need here). If I alsi use event.preventDefault in the touchstart event, scrolling does not work anymore.

So long story short: I don't really understand why it works, but it does. I testet this in Chrome 50 on Linux (normal and Touch-Device-Simulation mode), Midori browser (which pretends to be touch-enabled on normal PCs but only fires mouseevents) and an Android Tablet with Chrome 49. Unfornunately, I'm have no iPad to test.

ChristianMayer:
It works better than the previous pull request.

But, it still changes functionality on the Galaxy Note 10.1 and an iPad in comparison to the last release:
On the release it's possible to cancel a button press by keeping the finger down and dragging it out of the button. With this pull request that's not possible anymore.

On the desktop Chrome by using the mouse it's possible to do this cancelation, with release and this pull request.

peuter:
Ok next try, there was a leftover of the pointerevents now the listener is listening to the the touchmove event and the old behaviour works on my test system.

PR #311, opened 2016-04-10
Feature/split jqueryui

peuter:


ChristianMayer:
Actually the version that I've updated yesterday does contain only the used parts. What benefit do you see in including everything into the repository now just to leave most of it unused?

peuter:
Oh I didn't realize that. So in this case the PR is obsolete.

PR #312, opened 2016-04-10
Update license header

peuter:
This PR updates the license header in the source files to a 'grunt usebanner' compatible version. From now on changes/updates to the license headers (e.g. updating the license year) could be done by a grunt task.

Informations from the current headers (author + year of creation) have been moved to @author + @since annotations in the file header.

Note: This PR is not so much about what the new license header contains (as we can change that later easily) but about preparing the automation without loosing information.

### DIFF
--- a/src/lib/pagepartshandler.js
+++ b/src/lib/pagepartshandler.js
@@ -1,6 +1,6 @@
 /* pagepartshandler.js 
  * 
- * copyright (c) 2010-2016 by Christian Mayer (ChristianMayer) [CometVisu at ChristianMayer dot de]
+ * copyright (c) 2010-2016, Christian Mayer and the CometVisu contributers.
  * 
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the Free
--- a/src/lib/pagepartshandler.js
+++ b/src/lib/pagepartshandler.js
@@ -18,7 +18,6 @@
  *
  * @module Pagepartshandler 
  * @title  CometVisu Pagepartshandler 
- * @version 0.9.1-RC2
  */
 
 
--- a/src/lib/pagepartshandler.js
+++ b/src/lib/pagepartshandler.js
@@ -21,6 +21,7 @@
  * @version 0.9.1-RC2
  */
 
+
 /**
  * @author Christian Mayer
  * @since 2010
--- a/src/lib/pagepartshandler.js
+++ b/src/lib/pagepartshandler.js
@@ -1,5 +1,7 @@
-/* pagepartshandler.js (c) 2010-2015 by Christian Mayer [CometVisu at ChristianMayer dot de]
- *
+/* pagepartshandler.js 
+ * 
+ * copyright (c) 2010-2016 by Christian Mayer (ChristianMayer) [CometVisu at ChristianMayer dot de]
+ * 
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the Free
  * Software Foundation; either version 3 of the License, or (at your option)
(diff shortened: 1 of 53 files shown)


src/lib/pagepartshandler.js
```diff
@@ -1,23 +1,30 @@
...
  * more details.
  *
  * You should have received a copy of the GNU General Public License along
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
- * 
- * @module PagePartsHandler
- * @title  CometVisu templateengine
+ *
+ * @module Pagepartshandler
```
ChristianMayer: Is CamelCase not allowed here? `PagePartsHandler` is written in CamelCase...

src/structure/pure/page.js
```diff
@@ -1,20 +1,30 @@
...
  * 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+ *
+ * @module Page 
+ * @title  CometVisu Page 
  */
 
+
+/**
+ * @author Christian Mayer
+ * @since 2012
```
ChristianMayer: Quite often 2012 is used in the files, but the content is older, most likely from 2010. If that's relevant I guess we need to dig a bit in the older SVN code...

src/structure/pure/refresh.js
```diff
@@ -1,3 +1,26 @@
+/* refresh.js
```
ChristianMayer: I seems the header I now doubled in this file

src/structure/pure/reload.js
```diff
@@ -1,3 +1,26 @@
+/* reload.js
```
ChristianMayer: also doubled

ChristianMayer:
OK, looking through the commit I see a few doubled headers now and quite often a @since 2012 although I'm pretty sure the code is older.
Also the top copyright line is now always stating 2010-2016, even when the file was created later. And it's stating me even when the file was created by someone else.

How shall we solve that? Do we need to solve all of it or is some of that acceptable? (I'm not a copyright expert...)

src/lib/pagepartshandler.js
peuter: It is allowed, but the module part is autogenerated by grunt from the filename with first-letter-uppercase. In order to have CamelCase here the filename must be CamelCased too (as I think about wouldn't that be better option to have filename == modulename?)

src/structure/pure/page.js
peuter: The years are taken from the old headers. I don't think that we really need to take the effort to look into the sources.

peuter:
I removed the double headers (hopefully I got them all now). I think the since isn't really that important, to take the effort to search for the first commit of that file.

I am no copyright expert either.
I would see the header as the License/Copyright notice for the whole CometVisu project and not so much per file. There were files before, which just c&p the header and then the code was claiming you as creator even though you weren't (did that myself, too ;-).

I don't really know how to handle that the right way. Maybe we can say:

copyright (c) 2010-2016 by the CometVisu Project

If you are ok with that? But as the "CometVisu Project" is no legal entity I don't know if that is the correct way to go.

src/lib/pagepartshandler.js
ChristianMayer: Sounds good. I'd put that change into a separate PR though.

ChristianMayer:
After doing some research I cam across http://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/

According to that it would be:
> Copyright 2010 - [current year], Christian Mayer and the CometVisu contributors.

Personally I think that would be a good solution

Makki1:
I'm fine with that, Signed-off-by Michael Markstaller, means you can put anything of my code under a somewhat GPL-compatible license

peuter:
OK copyright line changed according to you suggestion

PR #313, opened 2016-04-10
update version

peuter:
This PR is meant to discuss how to proceed with versioning in the CometVisu. I would suggest to use the commonly used Semantic Versioning System (http://semver.org/), which is split into major/minor/bugfix versions. So the next feature version would be 0.10.0, and we would have a 0.9.2 if we release a bugfix only version.

As semver seems to be the de-factor-standard the javascript/grunt/npm world, I think it would better to break with the currently used versioning system and use that one instead, right?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "CometVisu",
   "description": "The CometVisu is an interactive, realtime, web based visualization. It can be used to access the KNX bus or with the OpenHAB backend everything in the OpenHAB world.",
-  "version": "0.9.1-RC2",
+  "version": "0.10.0-dev",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


ChristianMayer:
I thought the current naming scheme is already in accordance to the semantic versioning system (at least that's what I tried to achieve...).

A different question it the name of the next release. Due to the merge of the universal client I agree that the next release based on the current HEAD must be more than a bugfix release.

peuter:
As the 0.9.1 introduces the responsive layout as a new feature it is not a pure bugfix release IMHO. I would suggest, to always increase the minor version in the development branch after a release, because usually the next release introcudes new features but after the release they might not exist yet.

But for now we only have to decide about the next version, as you are fine with 0.10.0 I would suggest to merge this PR.

PR #318, opened 2016-04-23
Feature/coveralls

peuter:


ChristianMayer:
Why is the diff of the README so big? Are the line endings changed?

peuter:
No its the filename (suffix .md added) in order to tell github that this is a ile in markdown syntax.

PR #319, opened 2016-04-23
Fix DPT6.010

ChristianMayer:
and also make DPT5.004 more robust by ensuring the data range

This will fix https://knx-user-forum.de/forum/supportforen/cometvisu/940121-dpt-6-010

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "csso": "^1.5.4",
     "grunt": "latest",
     "grunt-banner": "^0.6.0",
-    "grunt-bump": "^0.7.0",
+    "grunt-bump": "0.7.0",
     "grunt-chmod": "latest",
     "grunt-contrib-clean": "^1.0.0",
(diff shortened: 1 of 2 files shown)


peuter:
Could you please also integrate the changes from commit 627015e746a08bccfa55c20622fbabdbeb9fc967 to fix the build either be also cherry-picking or as it is a tiny change by manually editing the package.json? Feels better to accept PR with working builds, altough in this version the build does not do that much :-)

ChristianMayer:
Picked it. But it seems it didn't help :(

BTW: please also cherry pick your commit that should go into the 0.9.2

peuter:
Ok seems as if this one is also needed 2b72ee375dada5c990a9dc2565a048dcc9843ab6

Will check which commit should be part of 0.9.2 tomorrow.

ChristianMayer:
Better now :)

PR #321, opened 2016-04-23
Fix editor after jQuery update

ChristianMayer:
To keep things simple this will only give the old jQuery version to the editor. Otherwise the editor would also need to be updated to use the new build system, RequireJS, ...
-> a new project on its own


peuter:
Just for curiosity: wouldn't it be possible to add the jquery.migrate plugin in the editor, to make the new jquery version compatible to the editor. The purpose would to not to have 2 different jquery versions in the project.
I did not check it myself, just read it here:http://api.jquery.com/jquery.browser/. So this is just a question if that would be an option?

ChristianMayer:
Generally: yes. And I used it temporarily during the CV update to the new jQ as it gives warnings on the console.

But in this case I wanted to fix the editor quickly without causing any unknown problems. So I decided to take a stable approach by using the well known jQ version.

PR #323, opened 2016-04-24
rename source files to use CamelCase syntax

peuter:
That topic came up in the discussion about the changed source header PR: To have module name == file name. So I tried out the refactoring functionality of the intellij ide and this is the result.


ChristianMayer:
I guess the other merged created a conflict here.

By looking on the diff I'd merge it, but I wanted to merge the "more important" other PRs first.

peuter:
Ok conflicts were solved and I think we can live the the coverage decreasing by 0.07% ;-), although I have no idea why that happended?

ChristianMayer:
I was also wondering about the 0.07% - probably one last linefeed or so

Issue #337, opened 2016-05-16
Refactor startup code


ChristianMayer:
The current code and the events called during startup "is grown" and should be thoroughly refactored and documented.

Special tasks to take care of:
- Handling of delayed resource availability (e.g. backdrop image)

ChristianMayer:
This should also consider when and how the handleResize is called.

ChristianMayer:
Pull request https://github.com/CometVisu/CometVisu/pull/342 addresses the handleResize. But it can be done more thoroughly

ChristianMayer:
This should be done during a big code cleanup -> post 0.10.0

ChristianMayer:
See also #121

peuter:
IMHO this has been done in 0.11. Can we close this issue?

peuter:
Done in 0.11

PR #341, opened 2016-05-26
Enhance 2d pages with contained mode

ChristianMayer:
The scaling option "contained" will scale the background to be as big as possible while still maintaining aspect ratio.
The elements with layout tags are scaled accordingly


peuter:
@ChristianMayer Could you provide an example (maybe in the demo config) where I can test the new feature?

And do you see any possibility to provide a test for this feature? I suggest to introduce a policy where we only accept new features, when the PR includes some kind of testing (not for plugins, but for core parts)

ChristianMayer:
@peuter Demo added, called demo_flat

Actually I wanted to add it from the beginning but forgot it. It's the same demo that I created for the Light & Building trade fair to show it on the ElaboratedNetworks demo WireGate and that was used as the screenshot for the 0.9.1 release. There it was just a static backdrop without this "contained mode" and the use of devices with highly varying resolution showed me the need to implement this feature.

Note: this demo is using a SVG backdrop. This works well on desktop browsers and on iOS. But my Android 4 tablet has a strange drawing bug. Here I'm pretty sure it's an Android bug and not a CV issue. When the demo is changed to a PNG backdrop (also provided in the pull request) Android is also working fine.

ChristianMayer:
> I suggest to introduce a policy where we only accept new features, when the PR includes some kind of testing (not for plugins, but for core parts)

Yes and no. I think we should discuss that more thoroughly on the Forum as it's an important topic.

peuter:
contained

When I maximize the window to fullscreen it looks broken (see screenshot). Maybe you can have a look at this, seems that there is a resize-handling missing somewhere.

ChristianMayer:
What browser are you using?

By filing the bug report for the Android browser (https://bugs.chromium.org/p/chromium/issues/detail?id=615556) I figured out that my desktop Chrome is also affected. But I've seen that only on the simple test case. The CV was always working.

I'll switch the demo from SVG to a PNG backdrop and that shouldn't be a problem anymore till the Chrome browsers are fixed

peuter:
I am using Chrome too. Now its better but there still is a lighter gray border on the bottom.
contained

ChristianMayer:
Strange, I can reproduce that in sporadic, rare instances.
I'm pretty sure it's based on the overloaded resize handling that really needs a good cleanup, see also https://github.com/CometVisu/CometVisu/issues/337

I suggest to merge this first as it's only adding functionality. The cleanup should then be a different pull request to separate the issues. It wouldn't make sense to include the cleanup in this request.

Issue #336, opened 2016-05-16
Icons should follow text color


ChristianMayer:
Icons that have no specific color given should have the same color as the current text.

ChristianMayer:
@peuter I'm thinking of changing the recolorable icons from a `<canvas>` to a `<div>` with the canvas as a background image to be hopefully able to use CSS as the deciding factor for the color used.

How could that be compatible with using a web font for the usual icons?

(Note: even when the web font is used for most of the icons the infrastructure for the icon recoloring is still necessary as the user could allways provide own icons that are not in the web font)

ChristianMayer:
See also https://github.com/CometVisu/CometVisu/issues/251

peuter:
Shouldn't be a problem, because using webfonts you only need a tag with some css to apply the webfont and the icon-character as text-content. In the current version it is done somehow like <i class="icon">[char]</i>. Could be easily changed to a div. I think is it even more compatible with webfont icons, as they are also colored by CSS.

ChristianMayer:
OK, then I'll try an implementation

ChristianMayer:
Hm, that also doesn't seem to work as intended. (I can make the color follow a class like "green" - but it doesn't follow the text color itself that might be changed somewhere. So it's also not (easily) possible for follow a text color change imposed e.g. by pressed vs. unpressed).

Next attempt: use SVG icons. https://sarasoueidan.com/blog/icon-fonts-to-svg/ ,  https://css-tricks.com/svg-sprites-use-better-icon-fonts/ and https://css-tricks.com/cascading-svg-fill-color/ et al. are talking about it beeing superior to icon fonts anyway...

PR #345, opened 2016-05-29
Switch icon to svg

ChristianMayer:
Make icons use a (very :() big SVG file instead of the many little PNG files.


ChristianMayer:
This pull request solves https://github.com/CometVisu/CometVisu/issues/336.

But note: the currently(?) generated SVG file is very big - 6.5 MB. Zipping it makes it more acceptable, but even then it's 1 MB big.

I'm not sure whether we want to give every user such a big impact when they are most likely using only a small subset of the icons.

peuter:
I would suggest to move this one to an extra branch in the main repo, as I did some tweaks when I tried to generate a webfont from the SVG Icons, which could be useful here too. Those tweaks fixed some problems and significantly decreased the size of each SVG (by 30-40% as far as I remember).

ChristianMayer:
OK, it's now in https://github.com/CometVisu/CometVisu/tree/switch_icon_to_svg

ChristianMayer:
Will be replaced by the PR on the CometVisu/CometVisu branch.

PR #349, opened 2016-06-10
login before generating page (setup_page() call): allows backends to …

peuter:
…send some configuration to the client, which might be needed during page setup (e.g. for the diagram plugin)

### DIFF
--- a/src/lib/CometVisuClient.js
+++ b/src/lib/CometVisuClient.js
@@ -612,12 +612,12 @@ define( ['jquery'], function( $ ) {
         this.stop(); // stop when new addresses are empty
       }
       else if (startCommunication) {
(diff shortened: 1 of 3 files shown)


joltcoke:
Excellent, both issues (the diagram and the initial value polling) are fixed! Thank you!

ChristianMayer:
@joltcoke did you review that PR thoroughly or check only that the fix is working?

I'm just asking as any change to lib/CometVisuClient.js must be considered as highly critical (and not as I expect a problem here)

joltcoke:
Yes, i did extensive testing, however i'm only having a specific setup, so my coverage for the whole codebase isn't close to 100%. 

I agree that this change is rather critical and also took a deep look at the changes.

PR #356, opened 2016-06-28
Update rsslog_external.php

ctr49:
prevent Cross-Origin problems as suggested in https://knx-user-forum.de/forum/supportforen/cometvisu/34494-rss-vs-rsslog?p=877559#post877559


ctr49:
also make call to feed API dynamic to fetch fresh content as suggested in #219

PR #358, opened 2016-07-05
fix pagejumps: use different key for target-path in data storage to s…

peuter:
…eparate it from the widget path


cool-runnings:
Hello,

I see that you merged the file into cometvisu/develop, but it is not present in the file.

peuter:
Hm somehow it got lost somewhere, strange. Anyway I just cherry-picked the commit and merged it directly as it has been already reviewed. Now its there.

PR #359, opened 2016-07-09
Fix/colorchooser

peuter:
fixes initial sending of data to the backend when colorchooser gets loaded.

see: https://knx-user-forum.de/forum/supportforen/cometvisu/963486-problem-milight-colorchooser-mit-oh-als-backend

Also there are some minor code styling changes due to linter errors shown

### DIFF
--- a/src/plugins/colorchooser/structure_plugin.js
+++ b/src/plugins/colorchooser/structure_plugin.js
@@ -65,30 +65,31 @@ define( ['structure_custom', 'css!plugins/colorchooser/farbtastic/farbtastic.css
           var br = data.bus_r;
           var bg = data.bus_g;
           var bb = data.bus_b;
+          var v;
           for( var addr in address )
           {
-            if( !(address[addr][1] & 2) ) continue; // skip when write flag not set
+            if( !(address[addr][1] & 2) ) { continue; } // skip when write flag not set
             switch( address[addr][2] )
             {
               case 'r':
-                var v = Transform[address[addr][0]].encode( r );
-                if( v != Transform[address[addr][0]].encode( br ) )
+                v = Transform[address[addr][0]].encode( r );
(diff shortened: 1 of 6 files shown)


src/plugins/colorchooser/structure_plugin.js
```diff
@@ -138,29 +141,29 @@ define( ['structure_custom', 'css!plugins/colorchooser/farbtastic/farbtastic.css
     switch( wData.address[ ga ][2] )
     {
       case 'r':
-        wData.bus_r = value;
+        wData.bus_r = value * 100 / 255.0;
```
ChristianMayer: Only looking at the code (not really trying it): this seems wrong.

Here you are assuming value in the range of [0...255] and map it to [0...100](%) - and in line 146 below you are assuming value to be in the range of [0...100] and map it to [0...255]

src/plugins/colorchooser/structure_plugin.js
```diff
@@ -158,9 +161,9 @@ define( ['structure_custom', 'css!plugins/colorchooser/farbtastic/farbtastic.css
                 color.substring(7);
         break;
       case 'rgb':
-        wData.bus_r = value[0];
-        wData.bus_g = value[1];
-        wData.bus_b = value[2];
+        wData.bus_r = value[0] * 100 / 255.0;
+        wData.bus_g = value[1] * 100 / 255.0;
+        wData.bus_b = value[2] * 100 / 255.0;
         color = color.substring(0,1) +
```
ChristianMayer: The same as above. Is that really correct?

By comparing the cases I'd expect that lines 168, 169 and 170 should have been changed to `toHex( value[...] * 255/100 )` instead

peuter:
You are right, I changed it.

PR #366, opened 2016-08-14
Api Doc including build system

peuter:
This PR introduces an automatic API-doc generation system, wich generates a html source code documenation including auto-generated example screenshots for widgets.

Although the api doc comments are not complete I create the PR now, as it is already a huge changeset and the missing widget docs are marked with a todo.

Mainly the changes in the code refer to missing indentation, parameter renaming and of course adding docs. But there is also a bugfix for the infotrigger (update didn't work at all + width and height setting was ignored) including changes to the unittest, which tested the wrong bevahiour before.

I also removed the @module and @title from all source code banners because module belongs to the doc part and title does not exist in the jsdoc specification.

I know that this one is hard to review due to its size, but splitting those things up would be even more work.


peuter:
btw. for testing the docs can be generated by 'grunt api-doc' (npm install must be done before that) and can be found in 'doc/api/html'

src/lib/CometVisuClient.js
```diff
@@ -15,21 +15,18 @@
...
- * @title CometVisu-Client
- * @exports ComentVisuVlient
- * @requires jQuery
+ * @module lib/CometVisuClient
+ * @exports ComentVisuClient
+ * @requires dependencies/jquery
  * @author Christan Mayer
  * @author Tobias Bräutigam
- * @since 2010
+ * @since 0.5.3 (initial contribution) 0.10.0 (major refactoring)
```
ChristianMayer: Sure? I thought the client was here since the beginning.

peuter: No not sure, I just took the first version after the year 2010 (as it was specified as the year of the contribution). Referring to the available releases on github I could say that is has been part of 0.5.1 (which is the oldest one there), I don't know/remember which one was the first public release. So any suggestions welcome.

src/structure/pure/ImageTrigger.js
```diff
@@ -15,100 +15,181 @@
...
+      var
+        $e = $(element),
+        classes = basicdesign.setWidgetLayout( $e, path );
+      classes += ' imagetrigger';
+      if( $e.attr('flavour') ) flavour = $e.attr('flavour');// sub design choice
+      if( flavour ) classes += ' flavour_' + flavour;
+      var ret_val = '<div class="widget clearfix image '+classes+'">';
+      ret_val += basicdesign.extractLabel( $e.find('label')[0], flavour );
+      var address = basicdesign.makeAddressList($e, null, path);
+      var width = $e.attr('width') || "100%";
```
ChristianMayer: It's really hard to do a review when multiple things are changed at the same time, like a feature enhancement in this documentation PR.

=> I'm just assuming now that this will work and are not explicitly checking it.

src/structure/pure/Include.js
```diff
@@ -15,32 +15,45 @@
...
+     * @method create
+     * @param {Element} element - DOM-Element
+     * @param {String} path - internal path of the widget
+     * @param {String} flavour - Flavour of the widget
+     * @param {String} type - Page type (2d, 3d, ...)
+     * @return {String} HTML code
+     */
+    create: function( page, path, flavour, type ) {
+      var p = $.get( $(page).attr('src') );
+      var p = $.ajax({
```
ChristianMayer: I know it's just following the orignial code, but having twice a `var p` seems odd.
=> Please add a `\\FIXME`

src/structure/pure/Line.js
```diff
@@ -15,31 +15,47 @@
...
+ * @widget_example <meta selector="hr">
+ *   <caption>A line tag which uses 50% of the screen size (6 of 12 available columns)</caption>
+ * </meta>
+ * <line><layout colspan="6"/></line>
+ *
+ * @module structure/pure/Line
+ * @requires structure/pure
  * @author Christian Mayer
- * @since 2012
+ * @since 0.8.0 (2012)
```
ChristianMayer: Hm, should this be the date / version of the file or of the widget?

The widget was here sice the beginning (cf. https://sourceforge.net/p/openautomation/code/HEAD/tree/CometVisu/branches/release_0.5.1/visu/visu_config.xsd ), the file was most likely created during the widget separation into a file per widget

src/structure/pure/Include.js
peuter: Yes this is what I mentioned on gitter about a month ago (and forgot about it then), will add a hint

src/structure/pure/Line.js
peuter: I would prefer to use the version the widget was introduced, as we might use this information for the user manual and we don't need the source file version there. But I don`t really know which widgets were there from the beginning (possibly all marked with 2012?)

ChristianMayer: I'd say the same. But let's pull this request as it is and fix the dates later

PR #367, opened 2016-08-15
New plugin for enertex smartmeter

ChristianMayer:


ChristianMayer:
To be discussed: is the plugin name "enertexsmartmeter" and the widget name "smartmeter" the best choice?

peuter:
Not sure about the naming, but would't it be better to use the same name for widget and plugin. And if so I would tend to use enertexsmartmeter as there might be other smartmeter plugins/widgets in the future and in that case smartmeter would be to generic.

Regarding the PR, I guess I will just merge it, because without hardware I cannot test it anyway.

ChristianMayer:
Same name: agreed (that came from the beginning where I thought I'd have multiple widgets in this plugin; but that's not the case / necessary anymore)

The name: what would you think of something like "spectrum" instead?

Screenshots: https://knx-user-forum.de/forum/supportforen/cometvisu/976667-neues-plugin-enertex-smartmeter-spektrum-anzeige

peuter:
If this is not vendor-specific and works with all hardware using that datapoint I would go for a more generic name, spectrum seems to be fine. But a little bit of documentation would be fine ;-), at least:

`
description

@module plugins/spectrum/structure_plugin
@requires structure/pure
@author Christian Mayer
@since 0.10.0
`

ChristianMayer:
Note: PR is not ready at the moment!

PR #370, opened 2016-08-22
User manual including toolchain for auto-generation

peuter:
and auto deployment in github pages


.doc/browser-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+    dpkg --extract /tmp/chrome.deb chrome-x
+    if [ $CV_VERSION = stable ]; then
+        mv chrome-x/opt/google/chrome chrome
+        mv chrome/google-chrome chrome/google-chrome-stable
+    else
+        mv chrome-x/opt/google/chrome-$CV_VERSION chrome
+        rm chrome/google-chrome
+    fi
+    ln -s google-chrome-$CV_VERSION chrome/google-chrome
+fi
```
ChristianMayer: Do we need that file like that - or should we first check whether a browser is already installed (most likely allways the case...)

peuter: There is no chrome installed in the travis images,so we have to install it.

PR #388, opened 2016-08-27
Handle XSD annotation references in the editor

ChristianMayer:


src/visu_config.xsd
```diff
@@ -156,13 +156,19 @@
...
   <xsd:attribute name="styling" type="xsd:string" />
+
   <xsd:attribute name="value" type="xsd:string" />
+
   <xsd:attribute name="flavour" type="xsd:string">
     <xsd:annotation>
-        <xsd:appinfo>level:expert</xsd:appinfo>
+      <xsd:appinfo>level:expert</xsd:appinfo>
+      <xsd:documentation xml:lang="en">Selection of a display variant. See also :ref:'flavour'.</xsd:documentation>
+      <xsd:documentation xml:lang="de">Auswahl der Darstellungsvariante. Siehe auch :ref:'Flavour flavour'.</xsd:documentation>
```
peuter: The correct RST-Syntax for titled references is be: :ref:`Flavour <flavour>` (surrounded by backticks and the reference by <>)

src/editor/lib/Schema.js
```diff
@@ -560,7 +570,26 @@ var SchemaAttribute = function (node, schema) {
...
+   * Transform documentation text to link to the online documentation when it
+   * contains a reference.
+   * 
+   * @return string The transformed input string.
+   */
+  function createDocumentationWebLinks( text )
+  {
+    return text.replace(new RegExp( ":ref:[`'](.+?)[`']", 'g'), function(match, contents){
+      var 
+        reference = contents.match( /^(.*?) *<([^ ]*)>$/ ),
```
peuter: Iḿ not really sure, but spaces might be allowed in the <>, so maybe the regex should allow anything but > like: ` /^(.*?) *<([^>]*)>$/`. What do you think?

ChristianMayer:
I couldn't find a syntax / spec for the labels. But now I'm assuming it can't contain a "<".

PR #395, opened 2016-08-29
Update documentation for widget "page"

SirTobiIV:
add German description for page name and add description for page attribute "visible".

### DIFF
--- a/src/visu_config.xsd
+++ b/src/visu_config.xsd
@@ -554,6 +554,7 @@
     <xsd:attribute name="name" type="xsd:string" use="required">
         <xsd:annotation>
             <xsd:documentation xml:lang="en">title of this page</xsd:documentation>
+            <xsd:documentation xml:lang="de">Titel der Seite</xsd:documentation>
         </xsd:annotation>
     </xsd:attribute>
     <xsd:attribute ref="flavour" use="optional" />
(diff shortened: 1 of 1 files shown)


src/visu_config.xsd
```diff
@@ -563,7 +564,12 @@
             <xsd:appinfo>level:expert</xsd:appinfo>
         </xsd:annotation>
     </xsd:attribute>
-    <xsd:attribute name="visible" type="xsd:boolean" />
+    <xsd:attribute name="visible" type="xsd:boolean">
+    	<xsd:annotation>
+        	<xsd:documentation xml:lang="en">Displays the widget (page name) which links to be page</xsd:documentation>
```
peuter: I would suggest "Show a widget, which links to the page" I don't think that it is clear what the text in the brackets means, and I think we do not need it here.

src/visu_config.xsd
```diff
@@ -563,7 +564,12 @@
...
+        	<xsd:documentation xml:lang="de">Zeigt ein Widget an (Seitenname), das auf die Seite verlinkt</xsd:documentation>
```
peuter: According to the above suggestion: "Zeigt ein Widget an, das auf die Seite verlinkt"

ChristianMayer: Oder: Zeigt ein Widget mit dem Seitennamen als Link auf die Seite.

src/visu_config.xsd
```diff
@@ -566,8 +566,8 @@
     </xsd:attribute>
     <xsd:attribute name="visible" type="xsd:boolean">
     	<xsd:annotation>
-        	<xsd:documentation xml:lang="en">Displays the widget (page name) which links to be page</xsd:documentation>
-        	<xsd:documentation xml:lang="de">Zeigt ein Widget an (Seitenname), das auf die Seite verlinkt</xsd:documentation>
+        	<xsd:documentation xml:lang="en">Displays a widget with the page name being a link to be page</xsd:documentation>
```
peuter: type: please change "to be page" to "to the page"

SirTobiIV:
I ahve changed that according to your comments.

peuter:
There is still a misspelled word in the english text, as I mentioned lately. If you fix that I would merge this pull request.

SirTobiIV:
done - sorry overlooked that comment

PR #413, opened 2016-09-08
auto-create the redirect structure from document source

peuter:
can be used as a datasource for https://github.com/CometVisu/cometvisu.github.io/blob/master/_createRedirectStructure.sh

the new feature is, that we can define in the documents source which old wiki pages it will replace. This data is read during build and extracted to a dedicated file in .doc/redirect-structure.sh. I assume the replacement map is easier to maintain this way. @ChristianMayer But you have to find a way to retrieve this filefrom the CometVisu repo, when you build the CometVisu Homepage. No idea how that can be done, fallback would be to copy the content manually every now and then.


ChristianMayer:
Where does the mapping information end up now?

Passing a path parameter to the redirect generation script should be easy.

peuter:
Currently is manually commited to the main repo: https://github.com/CometVisu/CometVisu/blob/develop/.doc/redirect-structure.sh

But maybe it should be added to the gh-pages branch, because then it can be auto-commited by travis. Can you add an URL to the redirect generation script? In that case it would be easy. Otherwise it would have to be a single-file checkout or direct download from the github-repo.

ChristianMayer:
Getting the data live over HTTP during the script run should be possible.

Issue #416, opened 2016-09-10
Documentation improvement


peuter:
Basics:
- [ ] Transfer missing content from wiki (+ add `@since` information to widget pages)
- [ ] Complete documentation colaboration part, describe scaffolding
- [ ] Extend scaffolding to plugins
- [ ] Fill the gaps, work through the TODOs: http://test.cometvisu.org/CometVisu/de/manual/colab/todos.html
- [ ] refactor structure to `/<language>/<version>/(manual|api|...)`

New content:
- [ ] Start english documentation (mix content from wiki + from widget api doc)
- [ ] add some openHAB1/2 specific documentation (installation, configuration, diagrams)
- [ ] 

Nice to have:
- [ ] shared design template for all documentation parts (manual, api) + homepage
- [ ] add version links (like in read-the-docs)
- [ ] improve documentation helper:
  - [ ] use virtualenc + setuptools
  - [ ] more assistance
  - [ ] add simple git repo auto-setup
  - [ ] add git workflow (start doc: update local repo, create new branch, save doc: commit + push, end doc: create pull request)

ChristianMayer:
Any (which?) points that should be done before the 0.10.0 release? 
(I.e. which points would block a 0.10.0 release?)

peuter:
IMHO There is no showstopper here, of course the documentation is not complete, buts its more than the old one, so not blocking a release.

JonasArnold:
In Version 0.10.2 the documentation of the startpage parameterr is wrong. With 0.10.1 the startpage URL parameter was changed so that the name of the page can be used instead of the ID. But the 0.10.2 manual still describes the old way (to use the ID and how to get it).

peuter:
Well the old way should still work. So its not wrong, just incomplete.

JonasArnold:
Actually the old way didn't work for me because my browser does not display the target URL of the Navbar-element (the `javascript:templateEngine.scrollToPage('id_0_33')` thing). Maybe this is a problem with my browser..

peuter:
That way might indeed be wrong described. You can just open the Developer Tools of your Browser and look into the DOM-Tree. There you can find all ids.

PR #419, opened 2016-09-10
improve declaration of mandatory elements and attributes

peuter:
refs #416


ChristianMayer:
Would it be possible to change the space to a no breaking space (&nbsp;) to avoid these line breaks?
linebreak

PR #433, opened 2016-10-15
fix release version

peuter:
just tried to `grunt build` from the current develop branch and found out that the generated release version  didn't load. This fix solves the loading problem but as I found it more by luck than real knowledge there might be a better solution. @ChristianMayer Maybe you can have a look at it, because I am not really familiar with requirejs.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -236,7 +236,8 @@ module.exports = function(grunt) {
           buildCSS: false,
           paths: {
             'css-builder': '../build/css-builder',
-            'normalize': '../build/normalize'
+            'normalize': '../build/normalize',
+            'jquery': 'empty:'
           },
 
           modules: [
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -207,6 +207,7 @@ module.exports = function(grunt) {
         src: [
           'index.html',
           'visu_config.xsd',
+          'dependencies/jquery.js',
           'dependencies/require-2.1.15.min.js',
           'dependencies/css.js',
           'icon/*.png',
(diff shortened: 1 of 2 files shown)


peuter:
Unfortunately the page with the plantuml debian package is down and that breaks the build. So we have to wait if it comes back, because I did not find an alternative.

Gruntfile.js
```diff
@@ -236,7 +236,8 @@ module.exports = function(grunt) {
           buildCSS: false,
           paths: {
             'css-builder': '../build/css-builder',
-            'normalize': '../build/normalize'
+            'normalize': '../build/normalize',
+            'jquery': 'empty:'
```
ChristianMayer: This looks rather strange to me - but I'm not a Grunt expert. What does "empty:" do?

peuter: It's not a grunt thing but a requirejs config option: http://requirejs.org/docs/optimization.html#empty
But I have no idea why it solves the problem.

src/main.js
```diff
@@ -76,7 +76,7 @@ require.config({
 
 var templateEngine;
 require([
-  'jquery', 'TemplateEngine'
+  'jquery', 'lib/TemplateEngine'
```
ChristianMayer: This can work (I have to check the code), but usually the paths go into  the paths section of the configs and here only the name is used. As it's already there (see line 30 of this file) I guess I need to dig deeper here

peuter: Yep should work as configured, but it doesn't. Without this change the "compiled" release cannot include the TemplateEngine.

ChristianMayer:
The correct fix is #462 - so closing this pull request

PR #439, opened 2016-10-21
add simple text to speech plugin

peuter:
currently chrome + firefox supports this. Should be tested with chrome as the german voice in firefox sounds like crap.

To test, add this to your config:

```
<speech lang="de">
  <address transform="OH:string" mode="read">Speak</address>
</speech>
```

Load the config with url param ?testMode=1 and execute `templateEngine.visu.write("Speak", "Hallo hier ist die Cometvisu")` in the browsers javascript console.

### DIFF
--- a/src/lib/CometVisuClient.js
+++ b/src/lib/CometVisuClient.js
@@ -175,9 +175,7 @@ define( ['jquery'], function( $ ) {
             this.readResendHeaderValues();
             session.update(data);
             this.retryCounter = 0;
-            if (!session.dataReceived) {
-              session.dataReceived = true;
-            }
+            session.dataReceived = true;
           }
 
           if (self.running) { // keep the requests going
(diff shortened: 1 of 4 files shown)


src/lib/CometVisuClient.js
```diff
@@ -175,6 +175,9 @@ define( ['jquery'], function( $ ) {
             this.readResendHeaderValues();
             session.update(data);
             this.retryCounter = 0;
+            if (!session.dataReceived) {
```
ChristianMayer: Do we need here (and in the places below) really the if statement? Wouldn't it be sufficient to just set `session.dataReceived` unconditionally?

ChristianMayer:
For the ToDo list:

- Transform is missing - this will cause the readout of pure hex numbers when the eibd backend is used
- sometimes repeated messages are desired - probably the current check should be enhanced by a timeout
- It would be great to be able to use a mapping, i.e. only mesage numbers are sent over the bus and the mapping translates that to the message to read

peuter:
Could you open an enhancement issue for this, I think it might get lost when its only attached to a closed pull request.

PR #440, opened 2016-10-22
introducing automatic english doc generation...

peuter:
... from source code comments for plugins and widgets. Add the basic doc structure and add some autogenerated content.


doc/manual/en/config/widgets/plugins/openweathermap/index.rst
```diff
@@ -0,0 +1,104 @@
...
+=========================
+
+.. api-doc:: openweathermap
+
+Description
+-----------
+
+.. ###START-WIDGET-DESCRIPTION### Please do not change the following content. Changes will be overwritten
+
+structure_plugin.js (c) 2015 by Stefan Borchert (stefan@borchert.cc)
```
ChristianMayer: I think the auto import made a mistake here

doc/manual/en/config/widgets/plugins/strftime/index.rst
```diff
@@ -0,0 +1,93 @@
...
+.. api-doc:: strftime
+
+Description
+-----------
+
+.. ###START-WIDGET-DESCRIPTION### Please do not change the following content. Changes will be overwritten
+
+This plugins integrates formated date and clock strings into based on strftime.
+
+Thanks to Michael Markstaller for implementing the jqclock plugin as reference.
```
ChristianMayer: I don't think that this line belongs in a manual.

When we take that line out of the source file would that automatically be fixed here or would that require a manual process call?

doc/manual/en/config/widgets/unknown/index.rst
```diff
@@ -0,0 +1,17 @@
...
+
+.. api-doc:: Unknown
+
+Description
+-----------
+
+.. ###START-WIDGET-DESCRIPTION### Please do not change the following content. Changes will be overwritten
+
+Fallback widget shown when an unknown widget is defined in the configuration.
+You must not use this one directly.
```
ChristianMayer: Can we suppress the docu generation for this widget?

It's just an error fallback.

src/editor/lib/DocumentationMapping.json
```diff
@@ -1,40 +1,4 @@
 {
   "_base": "http://test.cometvisu.org/CometVisu/", 
-  "audio": "/latest/manual/config/widgets/audio/index.html#audio",
```
ChristianMayer: Why are all these lines deleted?

doc/manual/en/config/widgets/plugins/strftime/index.rst
peuter: Actually the content between `.. ###START-WIDGET-DESCRIPTION###` and `.. ###END-WIDGET-DESCRIPTION###` gets replaced on every build with the content taken from the source. So removing that from source should be fine.

src/editor/lib/DocumentationMapping.json
peuter: By accident, didn't meant to commit this change. Will revert that.

src/plugins/gauge/structure_plugin.js
```diff
@@ -16,29 +16,7 @@
  */
  
 /**
- * This plugins integrates Gauge (in canavas) into the visualization
- *
- *
- * short documentation
- *
- * attributes:
```
peuter: This information is totally outdated, thats why I removed it.

peuter:
See commits ;-)

PR #443, opened 2016-10-23
Speech-plugin enhancements:

### LINKED ISSUE #442: Speech plugin
For the ToDo list:

- Transform is missing - this will cause the readout of pure hex numbers when the eibd backend is used
- sometimes repeated messages are desired - probably the current check should be enhanced by a timeout
- It would be great to be able to use a mapping, i.e. only message numbers are sent over the bus and the mapping translates that to the message to read

peuter:
- defineable timeout for repeating messages
- add transforms + mapping
- add german manual entry
- update source docs

closes #442


src/plugins/speech/structure_plugin.js
```diff
@@ -94,6 +147,7 @@ define( ['structure_custom' ], function( VisuDesign_Custom ) {
         selectedVoice = defaultVoice;
       }
       utterThis.voice = selectedVoice;
+      console.log("saying '%s' in voice %s", text, selectedVoice.name);
```
ChristianMayer: Are these `console.log` messages here on purpose (e.g. during the development and stabilizing phase) or were they forgotten to remove?

peuter: Yes on purpose, should be kept till its more stable. Would be nice to have some sort of logger, where you can define a log-level, then we could leave those debug-messages where they are and set the log level to INFO for a release, whichs skips those messages.

PR #445, opened 2016-10-23
add caching support for rendered DOM including all data needed

peuter:


ChristianMayer:
Trying the code the page `http://wiregate/visu_git/?config=demo&enableCache=1` doesn't load, the console shows the error message:
```
ConfigCache.js:87 Uncaught TypeError: (intermediate value).serializeToString(...).hashCode is not a function(…)
ConfigCache.toHash	@	ConfigCache.js:87
ConfigCache.isValid	@	ConfigCache.js:80
success	@	main.js:156
fire	@	jquery.js:3187
fireWith	@	jquery.js:3317
done	@	jquery.js:8785
(anonymous function)	@	jquery.js:9151
```
When I run the debugger `(new XMLSerializer()).serializeToString(xml)` is a string with the config file. So it's no wonder to me that `.hashCode()` can't be called.

src/lib/ConfigCache.js
```diff
@@ -0,0 +1,104 @@
...
+     * Returns true if there is an existing cache for the current config file
+     */
+    this.isCached = function() {
+      return localStorage.getItem(templateEngine.configSuffix + "." + this._cacheKey) !== null;
+    };
+
+    this.isValid = function(xml) {
+      // cache the result, as the config stays the same until next reload
+      if (this._valid === null) {
+        console.log("Current hash: '%s', cached hash: '%s'", this.toHash(xml), this.getData("hash"));
```
ChristianMayer: Shall this be kept in the next release or is it just for debug?

When it's the later, please give it a comment so that everybody knows that this line can be deleted before the next release

src/lib/MessageBroker.js
```diff
@@ -0,0 +1,88 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+ */
+
+/**
+ * Global MessageBroker for publishing/subscribing messages on defined topics
```
ChristianMayer: Can you give more informations what this does and when it should be used, as it looks like generic infrastructure code to me

src/lib/MessageBroker.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+  return {
+    // simulate a singleton
+    getInstance : function() {
+      if (!instance) {
+        instance = new MessageBroker();
+      }
+      return instance;
+    }
+  };
```
ChristianMayer: I guess this can be simplified to (untested)
```
  return (
    // simulate a singleton
    function() {
      if (!instance) {
        instance = new MessageBroker();
      }
      return instance;
    )()
```
so everywhere else `MessageBroker` can be used directly, and it's not necessary to call `MessageBroker.getInstance()`

src/lib/PageHandler.js
```diff
@@ -39,31 +39,24 @@ define([ 'jquery' ], function( $ ) {
...
-      currentPath !== '' && templateEngine.callbacks[currentPath].exitingPageChange.forEach( function( callback ){
-        callback( currentPath, target );
-      });      
-      
-      var 
-        page = $('#' + target),
-        callbacks = templateEngine.callbacks[target];
-      
+
+      currentPath !== '' && MessageBroker.getInstance().publish("path."+currentPath+".exitingPageChange", currentPath, target);
```
ChristianMayer: See comment in `src/lib/MessageBroker.js` how this line might be simplified to
`currentPath !== '' && MessageBroker.publish("path."+currentPath+".exitingPageChange", currentPath, target);`

src/lib/MessageBroker.js
```diff
@@ -0,0 +1,88 @@
...
+      }
+      this.__registry[topic].push([callback || this, context, priority || 0, once || false]);
+      // sort by priority
+      this.__registry[topic].sort(function(a, b) {
+        return b[2] - a[2];
+      })
+    };
+
+    this.publish = function (topic) {
+      if (this.__registry[topic]) {
```
ChristianMayer: The multiple calls to `this.__registry[topic]` can be optimized by puting it into a local variable.
(That 's a common optmization for JS and quite cheap due to the JS refence copy of arrays)

src/plugins/colorchooser/structure_plugin.js
```diff
@@ -56,85 +56,89 @@ define( ['structure_custom', 'css!plugins/colorchooser/farbtastic/farbtastic.css
...
-          var br = data.bus_r;
-          var bg = data.bus_g;
-          var bb = data.bus_b;
-          var v;
-          for( var addr in address )
-          {
-            if( !(address[addr][1] & 2) ) { continue; } // skip when write flag not set
-            switch( address[addr][2] )
+
+    this.construct(path);
```
ChristianMayer: What is the reason to create an extra `construct` method?

src/plugins/powerspectrum/structure_plugin.js
```diff
@@ -204,43 +204,48 @@ define( ['structure_custom',
...
           $e.attr('color2')     || "#cb4b4b",
           $e.attr('color3')     || "#4da74d"
-        ];
-      
+        ]
+      });
+
       // create the actor
       var actor = '<div class="actor clickable">';
       if( showCurve )
```
ChristianMayer: This will fail as `showCurve` is not defined anymore

src/lib/MessageBroker.js
peuter: I would prefer to explicitely use the Singleton-pattern here and leave it the way it is now.

src/lib/PageHandler.js
peuter: see above

ChristianMayer:
Bug :bug:: The (Inline) Diagrams are shown only as "loading...". Clicking on them opens correctly the big window with the diagram, but the inline is broken

ChristianMayer:
Bug :bug:: (Dunno if this PR caused it): the design toggle doesn't work anymore. It's empty and a click on it gives in the console the error message:
`DesignToggle.js:87 Uncaught TypeError: Cannot read property 'indexOf' of undefined(…)`

ChristianMayer:
Bun in templateEngine.js, line 111:
`this.enableCache = $.getUrlVar('enableCache') ? !!$.getUrlVar('enableCache') : true;`
this will allways stay true, even when the URL contains `enableCache=0` as `!!"0" === true`

BTW: I'd also replace the "0" and "1" by "true" and "false" as that's what the other URL parameters are using.

ChristianMayer:
Bug: In the demo config the footer isn't set up correctly.
cv

ChristianMayer:
Bug `?config=demorss`: many logs are missing

ChristianMayer:
Bug `/?config=gauge`: first start (no cache): gauges are not updating; cached start: gauges are not shown at all

ChristianMayer:
Bug `config=smartmeter` is only displaying w/o cache

peuter:
I cannot test the rsslog part. I think the php-Part is not compatible with PHP-Versions >= 5.5.0 because it uses `sqlite_open` which is only available <= 5.4.0. I can't get it running here and I don't want to rewrite the plugin just to test this PR.

And 2. I dont have a smartmeter config, so I am not able to test that bug.

peuter:
@ChristianMayer There still is a bug in the diagram plugin when cache is used, the other bugs (despite the two i cannot test) should be fixed.

ChristianMayer:
New Bug: Demo-Config. First (uncached) load is ok, second load shows corrupted layout and in the console the error message
`jquery.flot.min.js:1 Uncaught Error: Invalid dimensions for plot, width = 94, height = 0(…)`

PR #446, opened 2016-10-31
generate CVElements.yml from docs

peuter:


utils/config.ini
```diff
@@ -5,6 +5,8 @@ manual-templates=doc/manual/_templates
 cachedir=cache/
 schema-file=src/visu_config.xsd
 develop-version-mapping=latest
+www=http://test.cometvisu.org/CometVisu
```
ChristianMayer: Instead of putting the domain everywhere I'd prefer to have a relative link and make it absolute in the homepage files.

=> Can you switch the link generation to have the same "relativeness" as the manual itself?

peuter:
Done

PR #447, opened 2016-11-03
add Makefile to allow `make html`...

peuter:
... which is required by visual studio code to enable the live preview (shows the generated HTML-File, while you are editing the RST-File). 

This might be a nice feature for users that want to contribute to the documentation, but have no experience with the RST-Syntax.


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "restructuredtext.confPath" : "doc/manual/de/conf.py",
+    "restructuredtext.builtDocumentationPath" : "out/de/0.10.0-dev/manual/",
```
ChristianMayer: Must this be hardcoded? 
If so, a generic path should be used

peuter: Initilially I thought about automatically updating this path when the version changes (via our build tool), but a dedicated path just for this preview is indeed the better option, as it it version independent and does not has to be changed ever.

PR #456, opened 2017-01-06
Enhance manager for media

ChristianMayer:


peuter:
What is this about?

ChristianMayer:
An enhancement to fix https://knx-user-forum.de/forum/supportforen/wiregate/1030038-upload-von-bilder-f%C3%BCr-visu

PR #464, opened 2017-01-14
Fix counter image

ChristianMayer:


peuter:
Sorry guess I did something wrong, while testing this PR. After pulling to the last develop-Version the navbar in the Demo-Config looks kind of broken. Like the Pagejumps are floating when they should not. Could you check if this is wanted
bildschirmfoto vom 2017-01-15 09-45-52

ChristianMayer:
@peuter Please try with `enableCache=invalid` (like http://wiregate/visu_git/?config=demo&forceReload=true&enableCache=invalid ) - that solved it for me

I think we need to have a closed look at the Cache logic before the release...

PR #479, opened 2017-01-21
read config as argument from stdin when called via php-cli

ctr49:


ChristianMayer:
@peuter you might cherry pick this PR

PR #483, opened 2017-01-23
fix cached timer plugin

peuter:


ChristianMayer:
Reviewed by code inspection only

PR #485, opened 2017-01-23
- fix timeout plugin in cache mode (tested)

peuter:
- fix openweathermap plugin in cache mode (untested because of missing information how to use it)


ChristianMayer:
Reviewed by code inspection only

PR #492, opened 2017-02-05
Fix build system and demo config

ChristianMayer:


peuter:
Is there any difference to #493 or are they duplicates?

ChristianMayer:
Not duplicate, they have different targets

peuter:
Ah, ok, didn't realize that.

PR #517, opened 2017-02-20
add relative path hint for include widget

peuter:


.travis.yml
```diff
@@ -13,6 +13,7 @@ env:
   - CV_BUILD="test"
   - CV_BUILD="docs" ENCRYPTION_LABEL="4c79fec0aeb9" COMMIT_AUTHOR_EMAIL="generator@cometvisu.org" CV_BROWSER=Chrome CV_VERSION=stable
 before_install:
+- sudo apt-get update
```
ChristianMayer: I think this doesn't belong into this PR

peuter: Without this, the build fails. Of course I could create another PR for this, but I don't think its worth the effort.

peuter: Seems as if travis has fixed the probem, so I removed the apt update from the PR

PR #539, opened 2017-04-06
check if tags is array otherwise treat is as string

peuter:
this fixes the rsslog


ctr49:
confirmed to fix the problem, see [KNXUF](https://knx-user-forum.de/forum/supportforen/cometvisu/34494-rss-vs-rsslog?p=1080752#post1080752)

PR #540, opened 2017-04-06
iPad 1 (IOS 5.1.1) support

wuschi:
Some changes that make CometVisu work on iPad 1 (IOS 5.1.1) again:

* Added some workarounds for making remove() and bind() work on IOS 5.1.1 Safari
* Added SVG For Everybody (https://github.com/jonathantneal/svg4everybody) for supporting SVG External Content on older browsers


peuter:
As we are about to merge a major rework of the CometVisu based on the Qooxdoo-Framework, these changes would not work with the new version. Parts of this PR would be obsolete as Qooxdoo has lots of polyfills to support older browsers, e.g. I know that the bind-polyfill is part of the framework. Not sure about the remove one but it is very likely that is also exists.

The only leftover if the svg2everybody part. But that has to be included in another way because the new version does not use requirejs anymore.

So I suggest not to merge this PR, but to keep it open as a reminder to get back to after the new CometVisu version has been merged.

@wuschi It would be great if you could test if the polyfills are still needed for ios 5.1 (as I do not own such a system to test is myself) with the new version. Either after it got merged or right now from here:
https://github.com/peuter/CometVisu/tree/next

wuschi:
OK, i will check your branch and report back with the results when i have some time.

peuter:
Is has been merged now, so you can use the main repositories develop branch to test it. Svg2everybody is already included. It would be great if you could find the time to test, if it is working with yout iPad 1

wuschi:
Hi, i just did a quick test on my ipad. Unfortunately it doesn't work, at least with my custom visu_config.xml. The script console just logs an error "undefined - TypeError: 'undefined' is not a function". Debugging on IOS 5 is a pita, so i have to investigate this further by adding some console logging. The error does not show up with the default visu_config.xml, but nonetheless the text is not displaying.

peuter:
Did you try to load the config in 0.11.0 on a normal desktop? If the error is not device specific you might consider to record a log (see: http://cometvisu.org/CometVisu/de/latest/manual/colab/index.html#fehlerberichte-mit-log-dateien) and send that file to me. But if the error only occurs on your iPad1 the Log would not help.

wuschi:
The config loads fine on my desktop (Chrome on Mac). So I guess I have to trace the error on my iPad.

peuter:
As this is an issue I suggest to continue this discussion in issue #547 . Closing this PR

PR #548, opened 2017-04-15
add license header + backwards compability to client

peuter:


ChristianMayer:
What issue is fixed by this patch?

peuter:
This allows the client to be created with `new CometVisuClient()` in standalone version, as it was before. Without it you would need to use `new cv.io.Client()`.

PR #562, opened 2017-04-16
Fix slide widget on knxd backend

ChristianMayer:
As the encoded value is a (hex) string a NaN check doesn't work properly.
This patch fixes the slugish behaviour of the live updates.

Cf. #532 issue 1

### DIFF
--- a/source/class/cv/ui/structure/pure/Slide.js
+++ b/source/class/cv/ui/structure/pure/Slide.js
@@ -190,7 +190,7 @@ qx.Class.define('cv.ui.structure.pure.Slide', {
         var currentValue = this.getValue();
         this.sendToBackend(value, function(addr) {
           var newValue = cv.Transform.encode(addr[0], value);
-          return newValue !== cv.Transform.encode(addr[0], currentValue);
+          return newValue !== NaN && newValue !== cv.Transform.encode(addr[0], currentValue);
         });
       }
       this.setValue(value);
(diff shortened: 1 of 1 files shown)


peuter:
Well the transform encoding for openHAB backend might return `NaN` and this case should prevent sending that to the backend. As `isNaN(x)` returns false on strings, numbers etc. I dont see a problem here. It only returns true if x is actually `NaN`, so the hex string should work fine here. Could you provide an example value which breaks the current code? Maybe I'm getting something wrong here.

ChristianMayer:
Inserting in line 193: 
```
console.log( newValue, cv.Transform.encode(addr[0], currentValue), isNaN(newValue) );
```
gives e.g. during testing the result:
```
Slide.js:193 8062 8062 false
Slide.js:193 80ea 80ea true
Slide.js:193 8074 8074 false
Slide.js:193 8005aa 800352 true
```
The returned dataype of `cv.Transform.encode` is a string which IMHO makes sense as the communication over the network is always a string and not binary representation of numbers.
So from my perspective a `isNaN()` can't make sense.

The test here is to figure out whether the new slider value is exactly the same as that known on the network - including all rounding effects. That's why the encoded value is compared and not the raw value.

peuter:
Ok got it. Wrong understanding what isNaN does from me. I wanted to do something like this `newValue !== NaN` which is not the same as `!isNaN(newValue)`, lesson learned ;-)

But still, as the openHAB-Number-Transform encoding just does `parseFloat()` it might return `NaN` as value, and the purpose of this check is to prevent something stupid beeing send to backend.

So my suggestion would be to replace `!isNaN(newValue)` with `newValue !== NaN` so catch such cases.

PR #580, opened 2017-05-14
use YQL service for external RSS feeds

### LINKED ISSUE #461: Remove zrssfeed from rss plugin and rsslog_external.php from rsslog plugin
Both rss plugins depend on Google feed API that was finally turned off on January 10th.

- rss plugin was primarily meant to display extern rss feeds (via now also non-working zrssfeed)

- rsslog plugin was primarily meant to display the rsslog database, but was later extended to also display external feeds (which doesnt work either anymore).

I suggest to rewrite external rss feed support using a native rss2json in Javascript, but having above still in place will confuse users. So we should remove it.

For discussion also see:
https://knx-user-forum.de/forum/supportforen/cometvisu/34494-rss-vs-rsslog?p=1039458#post1039458

peuter:
closes #461


ChristianMayer:
Should the rssdemo be change to show support for external RSS feeds again?

peuter:
There is no change in usage so everything should work as before

ChristianMayer:
OK, it might be related to the line https://github.com/CometVisu/CometVisu/blame/develop/source/resource/demo/visu_config_demorss.xml#L133 using `rss` instead of `rsslog`

=> Should that be changed now?

peuter:
Seems so and as the RSS plugin only shows external feeds (and does not work any more). I think we can completely delete the plugin, right?

PR #581, opened 2017-05-15
remove harcoded height style and move other page > .clearfix styles t…

### LINKED ISSUE #473: Design alaska broken
The alaska and alaska_slim designs are currently broken:
cv2

peuter:
…o CSS

 closes #473


ChristianMayer:
The `height:100%;` get's removed from the `style` but not added to the CSS.

Are you sure that it's not needed anymore?

peuter:
It has to be removed because its causing the error. It should use "auto-sizing" by the content, so we must not set a height explicitly.

PR #590, opened 2017-05-22
make diagram initialization more robust

### LINKED ISSUE #584: Reloading of visu fails
If you select a different page, the url gets appended by e.g. #id_6_. If you reload the whole visu using the browser reload button, the visu won't load again. Reproduced with Safari and Chrome.

peuter:
closes #584

### DIFF
--- a/source/class/cv/ui/PageHandler.js
+++ b/source/class/cv/ui/PageHandler.js
@@ -84,8 +84,6 @@ qx.Class.define('cv.ui.PageHandler', {
 
       templateEngine.currentPage = pageWidget;
 
-      qx.event.message.Bus.dispatchByName("path."+target+".duringPageChange", target);
-
       // update visibility of navbars, top-navigation, footer
       templateEngine.pagePartsHandler.updatePageParts( pageWidget, speed );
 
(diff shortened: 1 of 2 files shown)


source/class/cv/ui/PageHandler.js
```diff
@@ -84,8 +84,6 @@ qx.Class.define('cv.ui.PageHandler', {
 
       templateEngine.currentPage = pageWidget;
 
-      qx.event.message.Bus.dispatchByName("path."+target+".duringPageChange", target);
```
ChristianMayer: Isn't this a more generic infrastructure to allow widgets and plugins do preperation work to show as soon as the page is available?
=> I guess it should still be in the code

peuter: I deleted is for two reasons:

1. No one is using it
2. There is no "during" page change. Because Javascript is single-threaded you can only do something before the page changes or after it changed. If you count in animations we might find a use case for for doing something when the page gets visible (and the animation starts). You can already listen to an "changeVisible" event on a page object to capture that moment. 
But it might be arguable to introduce a new bus message for this, when there is a need for it.

So long story short. As soon as there is a need for it I'm open to adding another bus event, but currently its not needed.

PR #600, opened 2017-08-14
Create cache control to fix image refresh

ChristianMayer:
Create attribute `cachecontrol` for image widget to control how the CometVisu makes the browser reload the image during refresh.

This fixes the recent server setup change that happened at yr.no  where it doesn't tolerate the addition of a question mark and timestamp to the URL any more.


peuter:
@ChristianMayer Do you see any chance to add unit tests for this new feature?

ChristianMayer:
For `full` and `weak` it could be possible, for `force` I fail to imagine a possible unit test to check the behaviour

Issue #606, opened 2017-09-04
Button animation and reaction


ChristianMayer:
During "non conventional" interaction the buttons (esp. `switch`) doesn't behave as expected / being used to at major OSes (Windows, Linux, Android).

Actions to compare (esp. with mouse as well as with touch):

- Press button and hold, move out, lift
- Press button and hold, move out, move back in, lift
- same but with scroll in between
- same but ended with `Esc` key
- ...

peuter:
And what is the expected behaviour?
* Press button and hold, move out, lift
   => do not send value
* Press button and hold, move out, move back in, lift
   => send value
* same but with scroll in between
   => do not send value?
* same but ended with Esc key
   => do not send value?

Found an additional behaviour difference for triggers:
* Press button and hold (button down animation => ok) 
* move outside the button (button up animation => missing)
* move back inside the button (button down animation => missing)
* lift  (button up animation => ok)

will fix that too.

ChristianMayer:
The issue is mainly about the animations (and the reaction in the sense of the animation).
I will check to be more precise where I see problems with the current state of the code.

ChristianMayer:
Tested with mouse against current develop (i.e. not #610) with demo config:

- Bind Click to Widget Trigger - clicking on the Widget and not the Button does send but doesn't animate
- Trigger: click, move out, release: when leaving the button it should look unpressed again (it's even worse at the moment: releasing the mouse that button is still looking pressed)
- Any button: click, move out, move in, release: should send, but doesn't
- Trigger: click, move out, move in, release: should look click=pressed - move out=unpressed - move in=pressed, release=unpressed,  but doesn't
- Switch: click, scroll and thus move out, move in, release: should send value but doesn't
- Pushbutton: allways looks pressed, but with label "An" should look unpressed (and during click look pressed and show the label "Aus")
- Multitrigger: doesn't show which state is currently set (pressed). Doesn't animate

On mobile device: I can't comment at the moment as the basics aren't working.
When those are fixed it should be checked whether it feels correct. Especially when the user wants to scroll and hits a button (or bind click on widget) either on purpose (as the intention is to click and not to scroll) or on mistake (as the user wants to scroll and not to click). The correct way to handle such a situation can be seen on native apps (e.g. in the OS config menu)

ChristianMayer:
Might relate to / fix #475

peuter:
TODO-List:
[...]

PR #610, opened 2017-09-09
fix button animation

peuter:
this is a work-in-progress PR to solve the issues mentioned in #606. As currently #606 does not contain a complete description of the known errors/misbehaviours, this PR handles only the known problem.

Initially these are:
* fixes button animation for all widgets using the HasAnimatedButton mixin (see http://cometvisu.org/CometVisu/en/latest/api/#cv.ui.common.HasAnimatedButton for a complete list). To test the fix, follow these steps: mouse-down on e.g. a trigger, move outside the trigger (button state changes to unpressed), move inside the trigger (button state changes to pressed), mouse-up

refs #606


peuter:
closes #606

ChristianMayer:
Tests with Chrome on a desktop with a mouse (i.e. not a mobile device):
* GestureCore.js:210 is contaminating the console.log
* Pushbutton (at least on demo config) now looking always pressed
* Clicking on the slider (instead of dragging it) doesn't always update the value

peuter:
* console.log removed
* Pushbutton: working fine for me. Only looking pressed when I change the switch abote to 'An' or use the pushbutton itself

peuter:
I am also not able to reproduce the slider problem.

peuter:
Ok, I added the listener to the sliders parent widget. Should be easier to use now, as you do not have to click exactly on the slider range. I assume thats was the problem with your missing slider updates.

ChristianMayer:
confirming Console.log fix.
But the pushbutton and slider is still broken.
The idea to make the slider react to outside clicks can be a good choice when the design decides to make the sliding-bar very thin. But for "normal" designs it will cause more troubles as clicks (very) left or right of the slider will also move it.
The issue I was describing also aren't related to miss the active area as the handle is moving - but the new values aren't always sent to the bus.

=> I suggest that 4ccdf64 is reverted and the pushbutton is fixed before this PR is merged. The uncommunicative slider can then be researched and fixed somewhere else as it's not an animation issue

ChristianMayer:
I guess the pushbutton is a type issue (number vs. string):
grafik

peuter:
Ok, done.

PR #635, opened 2017-10-15
fix rsslog datetime value transformation + icon appearance in popup

### LINKED ISSUE #627: RSSLog plugin shows no date
(see: https://knx-user-forum.de/forum/supportforen/cometvisu/1133276-rückmeldung-zur-aktuellen-entwicklungsversion-nach-umstieg-auf-qooxdoo?p=1139906#post1139906)

`rsslog.php` seems only to be compatible with PHP < 5.4.0 which is rather outdated. So this should be fixed be someone using the wiregate, to be able to verify the error and test the fix.

And/Or make the PHP-part of the plugin compatible with PHP 7. Using PDO as abstraction layer might be an option here, but it needs to be verified that this is available on wiregate too

peuter:
closes #627

### DIFF
--- a/source/class/cv/plugins/RssLog.js
+++ b/source/class/cv/plugins/RssLog.js
@@ -498,7 +498,7 @@ qx.Class.define('cv.plugins.RssLog', {
     },
 
     _onTap: function(ev) {
-      var item = ev.getTarget();
+      var item = ev.getCurrentTarget();
 
       var id = qx.bom.element.Dataset.get(item, 'id');
       var mapping = qx.bom.element.Dataset.get(item, 'mapping');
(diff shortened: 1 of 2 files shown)


peuter:
@ChristianMayer If this fix does not work, please record a log for me. I can see what going wrong then.

ChristianMayer:
Log sent by mail

ChristianMayer:
Icons are looking fine now.

But there's still an issue: when you click on an entry (first click opens a pop up, I'm talking about the click in the popup) you can "confirm" an item. This will change the icon and strike through the text.
The display of a confirmed item is fine - but the confirmation itself doesn't work: only the strike is shown but the icon isn't updated and the confirmation isn't stored.
Looking at the network is seems that there's a problem by sending a wrong "number" for `u`: `rsslog.php?u=undefined&state=1`
(Note that's the case most of the time, one of the entries did work...)

peuter:
Ok, found and fixed.

PR #637, opened 2017-10-16
use docker image to create screenshots in travis

peuter:
use the new docker helper image to create the screenshots: hopefully this fixes the problem of screenhots in wrong design


peuter:
I need the credentials to publish it there.

PR #650, opened 2017-10-19
Fix mem leak

### LINKED ISSUE #640: Memory leaks
I'm using an iPad 2 to show the visu. When I started using CometVisu the browser was crashing once a week due to out of memory situations. It got a lot better when jQuery and other components in CV were uploaded to the lastest version. The iPad was surviving weeks without any OOM crash.

Since the migration to the qx framework, it takes minutes, sometimes only seconds for the visu to crash. Upgraded to latest git version but same behaviour.

My start page only consist of some buttons and a weather map (image), showing only this page without switching to different pages will let the visu run for weeks. But as soon as you switch to a page with a diagram, the visu will crash fairly soon. Usually within seconds.

If I open my temperature diagram in Google Chrome on the desktop, I can see a steady increase of memory consumption as well.

peuter:
next try to fix #640


peuter:
not working this way

PR #652, opened 2017-10-20
Fix memory leak

### LINKED ISSUE #640: Memory leaks

peuter:
last try to fix #640.

When I disable Flot-Diagram painting I can see memory increasing after Diagram Data has been received, and some time later, when the garbage collector kicks in, the memory decreases again. Normal behaviour.

With flot, I cannot see the memory decreasing. Overall this PR improves memory usage, but in the long run I think that memory will still increase permanently. Needs to be verified in real use case.

So from my point of view, I have done everything possible to fix this. The only thing left would be to replace Flot with something else. As the last Flot release is dated in 2014, that would be the way to go anyways. But we need a volunteer to do so and that wont be me.

### DIFF
--- a/source/class/cv/plugins/diagram/AbstractDiagram.js
+++ b/source/class/cv/plugins/diagram/AbstractDiagram.js
@@ -195,6 +195,9 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
         this.cache[ key ].waitingCallbacks.push( [ callback, callbackParameter ] );
 
         if( this.cache[ key ].waitingCallbacks.length === 1 ) {
+          if (this.cache[ key ].xhr) {
+            this.cache[ key ].xhr.dispose();
+          }
           var xhr = new qx.io.request.Xhr(url);
           xhr.set({
             accept: "application/json"
@@ -216,6 +219,7 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
               waitingCallback[0](this.cache[key].data, waitingCallback[1]);
             }, this);
             this.cache[key].waitingCallbacks.length = 0; // empty array)
+            rrddata = null;
           }, this);
 
           this.cache[ key ].xhr = xhr;
@@ -633,6 +637,8 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
             plot.setData(fulldata);
             plot.setupGrid();
             plot.draw();
+
+            loadedData = [];
           }
 
         }.bind(this));
--- a/source/class/cv/plugins/diagram/AbstractDiagram.js
+++ b/source/class/cv/plugins/diagram/AbstractDiagram.js
@@ -202,32 +202,32 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
           xhr.set({
             accept: "application/json"
           });
-          xhr.addListener("success", function( ev ) {
-            var rrddata = ev.getTarget().getResponse();
-            if (rrddata !== null) {
-              // calculate timestamp offset and scaling
-              var millisOffset = (rrd.offset ? rrd.offset * 1000 : 0);
-              for (var j = 0; j < rrddata.length; j++) {
-                rrddata[j][0] = rrddata[j][0] + millisOffset;
-                rrddata[j][1] = parseFloat(rrddata[j][1][rrd.dsIndex]) * rrd.scaling;
-              }
-            }
-            this.cache[key].data = rrddata;
-            this.cache[key].timestamp = Date.now();
-
-            this.cache[key].waitingCallbacks.forEach(function (waitingCallback) {
-              waitingCallback[0](this.cache[key].data, waitingCallback[1]);
-            }, this);
-            this.cache[key].waitingCallbacks.length = 0; // empty array)
-            rrddata = null;
-          }, this);
-
+          xhr.addListener("success", qx.lang.Function.curry(this._onSuccess, rrd, key), this);
           this.cache[ key ].xhr = xhr;
           xhr.send();
         }
       } else {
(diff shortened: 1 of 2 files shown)


source/class/cv/plugins/diagram/AbstractDiagram.js
```diff
@@ -195,35 +195,39 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
...
+      var rrddata = ev.getTarget().getResponse();
+      if (rrddata !== null) {
+        // calculate timestamp offset and scaling
+        var millisOffset = (rrd.offset ? rrd.offset * 1000 : 0);
+        for (var j = 0; j < rrddata.length; j++) {
+          rrddata[j][0] = rrddata[j][0] + millisOffset;
+          rrddata[j][1] = parseFloat(rrddata[j][1][rrd.dsIndex]) * rrd.scaling;
+        }
+      }
+      this.cache[key].data = rrddata;
```
ChristianMayer: This stores a part of the response in the diagram. Probably prevents the GC of the XHR object?
It also changes type in place in line 221 which is said to cause bad performance in JavaScript as it forces the engine to switch from internal compiler to interpreter.
=> Suggestion: make lines 220 and 221 fill a new array and assign that to `this.cache[key].data` in line 224

peuter: I disabled the cache several times, makes no difference. This was the first suspect I had, that caching is responsible for the leak. But it turned out, that it isn't.

I tried your second suggestion. I might got it wrong, but creating a new array is slower then the old code.

```
Misc.js:40 PROFILE -> parse rrd response 1: 0.275146484375ms
Misc.js:40 PROFILE -> parse rrd response 2: 0.314208984375ms
Misc.js:40 PROFILE -> parse rrd response 1: 0.2197265625ms
Misc.js:40 PROFILE -> parse rrd response 2: 0.2509765625ms
Misc.js:40 PROFILE -> parse rrd response 1: 0.740966796875ms
Misc.js:40 PROFILE -> parse rrd response 2: 0.868896484375ms
Misc.js:40 PROFILE -> parse rrd response 1: 0.149169921875ms
Misc.js:40 PROFILE -> parse rrd response 2: 0.167236328125ms
```
Number 1 is the old code, number 2 the new one. And here is the code snippet:
```
        cv.util.Misc.time('parse rrd response 1');
        // calculate timestamp offset and scaling
        var millisOffset = (rrd.offset ? rrd.offset * 1000 : 0);
        for (var j = 0; j < rrddata.length; j++) {
          rrddata[j][0] = rrddata[j][0] + millisOffset;
          rrddata[j][1] = parseFloat(rrddata[j][1][rrd.dsIndex]) * rrd.scaling;
        }
        cv.util.Misc.timeEnd('parse rrd response 1');

        cv.util.Misc.time('parse rrd response 2');
        var newRrd = []
        for (var j = 0; j < rrddata.length; j++) {
          newRrd[j] = [(rrddata[j][0] + millisOffset), (parseFloat(rrddata[j][1][rrd.dsIndex]) * rrd.scaling)]
        }
        rrddata = newRrd;
        cv.util.Misc.timeEnd('parse rrd response 2');
```

ChristianMayer: I don't know where the result collecting table has gone to... Anyway: https://jsperf.com/type-cast-in-place
=> Firefox: "external prereserved" is fastest, "In place" is 88% slower
=> Chrome: "external prereserved" is fastest, "In place" is 37% slower

Also interesting:
- Performance of `map` is a desaster - quite poor for being the official best solution for such a task and also strange as a naive implemtation of `map` would/should do exactly the same as "external prereserved"
- in Firefox "In place" is much slower than both "external" versions that are quite similar
- in Chrome the `new Array(testArray.length);` helps quite a lot to boost performance

Conclusion:
Performance is - as allways - highly depending on the current engine used. But both major engines are preferring "external prereserved". So I'd stick with that.

peuter: Ok, I switched the positions of the two array traversals and it turns out that the first one is always the fastest one. So this is no reliable way to compare performance. I changed the code according to the fastest jsperf version

ChristianMayer: ??? https://jsperf.com/type-cast-in-place/2 has only one test left, so that will always be fastet and slowest on the same time ;)

peuter:
Hm, I am not able to solve the conflict in advance. So if you are fine with this PR now, I will try to merge it manually.

source/class/cv/plugins/diagram/AbstractDiagram.js
peuter: Sorry I tried to add one test, but it always complained about missing values in required fields. Did never accept my tries to save the test. Seems that it saved something broken without even noticing me. Strange behaviour.

peuter: Tried again, the jsper GUI makes absolutely no sense to me, I'll give it up to add another test there. Deleting revisions doesn't seem to be possible either.

PR #668, opened 2017-10-27
- use watchdog only in long-polling as SSE has auto-reconnection hand…

peuter:
…ling by the browser and the watchdog breaks the native mechanism

- restart sse connection only when it is really closed
- stop watchog if its already running on start (prevent multiple running intervals)

### DIFF
--- a/client/source/class/cv/io/Client.js
+++ b/client/source/class/cv/io/Client.js
@@ -73,9 +73,6 @@ qx.Class.define('cv.io.Client', {
 
     this.backendUrl = backendUrl;
 
-    this.watchdog = new cv.io.Watchdog();
-    this.watchdog.setClient(this);
-
     this.addresses = [];
     this.initialAddresses = [];
     this.filters = [];
@@ -217,7 +214,6 @@ qx.Class.define('cv.io.Client', {
    ******************************************************
    */
   members: {
-    watchdog: null,
     backend: null,
     backendName: null,
     backendUrl: null,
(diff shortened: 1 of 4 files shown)


ChristianMayer:
Give's an error with LongPolling:
```
LongPolling.js:149 Uncaught TypeError: Cannot read property 'ping' of undefined
    at wrapper.handleRead (LongPolling.js:149)
    at wrapper.dispatchEvent (Direct.js:135)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js:537)
    at wrapper.dispatchEvent (Manager.js:887)
    at Object.__fireEvent (Registration.js:302)
    at Object.fireEvent (Registration.js:321)
    at wrapper.fireEvent (MEvent.js:181)
    at wrapper._fireStatefulEvent (AbstractRequest.js:826)
    at wrapper.__onReadyStateDone (AbstractRequest.js:748)
    at wrapper._onReadyStateChange (AbstractRequest.js:723)
    at construct.onreadystatechange (Function.js:232)
    at construct._emit (Xhr.js:515)
    at construct.__readyStateChange (Xhr.js:1008)
    at construct.__onNativeReadyStateChange (Xhr.js:984)
    at XMLHttpRequest.<anonymous> (Bootstrap.js:715)
```

peuter:
I had a merge conflict after a branch changing, some changes obviously got lost dureing conflict solving. Please try again.

ChristianMayer:
There are no more error messages when using LongPolling - but the watchdog is broken.

With no traffic: after 60 seconds comes a new `r` - as it should
But that's stopping then. After 5 minutes the server is terminating and the client is immediately reconnecting (as it is supposed to do in this case) - but not the 60 second interval that should be coming from the client.
This results in the problem that when the internet connection is removed, I do get on the console a (expected)
```
GET http://wiregate/cgi-bin/r?t=0&s=SESSION&a=12/9/2&a=7/4/51&a=17/4/51 net::ERR_INTERNET_DISCONNECTED
```
but the client never recovers from that. I.e. when the connection comes back the `r` is never started again.

PR #672, opened 2017-10-27
Immediately load value and switch to PHP version of the service script

ChristianMayer:


source/class/cv/ui/structure/pure/WgPluginInfo.js
```diff
@@ -51,15 +55,16 @@ qx.Class.define('cv.ui.structure.pure.WgPluginInfo', {
       if (value) {
         if (!this.__request) {
           // create the request
-          this.__request = new qx.io.request.Xhr('/wg-plugindb.pl?name=' + value);
+          this.__request = new qx.io.request.Xhr('/wg-plugindb.php?name=' + value);
           this.__request.set({
             accept: "application/json",
-            async: false
+            async: true
```
peuter: as async: true is the default you can just delete it here

source/class/cv/ui/structure/pure/WgPluginInfo.js
```diff
@@ -20,6 +20,10 @@
 
 /**
  * Adds an dynamic field to the visu that shows live information from a WireGate plugin.
+ * 
+ * Note: In the web root, i.e. the directory /var/www/ must be this service
+ * script "installed":
```
peuter: ... directory /var/www/ this service must be "installed"...

Sounds better to me, or not?

PR #674, opened 2017-10-28
Fix wg plugin info

ChristianMayer:


peuter:
Didn't I merge this change yesterday? Looks familiar.

ChristianMayer:
Yes it's merged already. I've confirmed that by looking at the latest code. So this PR is obsolete.

But it's strange that GitHub still shows a diff that looks as it wasn't merged...

PR #675, opened 2017-10-28
Fix slider not sending values 

### LINKED ISSUE #655: Slider not allways working
As described in #610 the slider doesn't always send it's new value when clicked on the slider bar (long polling backend, not sending confirmed by network pane).
When the handle is dragged the new values are mostly sent - but sometimes they are also missing.

peuter:
- fix throttle function parameters
- remove !isNaN

closes #655


ChristianMayer:
:+1: closes #655

PR #678, opened 2017-10-29
disable drag boundaries

### LINKED ISSUE #677: Slieder can't be pulled to the exact end
Dragging the slider handler (quickly) to the end will stop sending before reaching the min or max value.

Most likely this is caused by the slider ignoring mouse positions outside of the sliding range.
But it should be possible (as in older releases) when the handler is dragged left/right of it's moving range that it reaches exactly the min/max value and sends that to the bus.

peuter:
closes #677


peuter:
Please try again.

PR #682, opened 2017-11-01
Enhance sendToBackend

ChristianMayer:
Add optional third parameter currentBusValue that can be used to prevent retransmission of already known values on the bus.

Usecase are widgets that are continuusly sending like a slider.

### DIFF
--- a/source/class/cv/ui/common/Operate.js
+++ b/source/class/cv/ui/common/Operate.js
@@ -71,28 +71,33 @@ qx.Mixin.define("cv.ui.common.Operate", {
      *
      * @param value {var} value to send
      * @param filter {Function} optional filter function for addresses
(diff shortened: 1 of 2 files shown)


ChristianMayer:
It's in preparation for the ControllerInput. But to demonstrate its use I've switched the slider to use it now - you can see that it's code is getting more compact by this new feature.

This feature is helpful for all widgets and plugins that have a continuous sending but where the bus shouldn't be flooded by repeated messages. As you can see on the slide this can now be achieved with no additional code line on the widget side.
This feature must also make sure that intended repeats are not blocked. This it the case e.g. with the trigger widget. But by not passing `currentBusValues` parameter this additional feature is disabled and the value will be sent.

The issue with multiple sending values should be handled now.

Issue #681, opened 2017-11-01
Error on page reload with diagram visible


joltcoke:
If I reload the visu with a page containing a diagram currently visible, I get the following error:

Error: Invalid dimensions for plot, width = 0, height = 100
    at Canvas.resize (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:5:4449)
    at new Canvas (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:5:4271)
    at setupCanvases (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:5:21088)
    at new Plot (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:5:11459)
    at Function.$.plot (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:5:52561)
    at cJ.initDiagram (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:4:8729)
    at cJ. (http://<ip>:8080/cv/build/script/cv-webkit.a48b41ca9405.js:4:13422)
    at cJ.dispatchEvent (http://<ip>:8080/cv/build/script/cv-webkit.js:213:153117)
    at construct.dispatchEvent (http://<ip>:8080/cv/build/script/cv-webkit.js:213:139438)
    at Object.__de (http://<ip>:8080/cv/build/script/cv-webkit.js:213:141418)

Latest GIT version, reproduced with Safari and Chrom on OSX.

peuter:
This is a duplicate of #663

joltcoke:
I tend to disagree. #663 is (initially) a different issue. The symptoms raised here: https://github.com/CometVisu/CometVisu/issues/663#issuecomment-339579267 are similar, however according to your statement the problem should be fixed by #665. However the fix doesn't work in my case.

I'd vote to keep this open and track the diagram error here instead of mixing various problems in #663.

peuter:
However: I tried to reproduce the issue with Chrome 62 (LInux) and Safari 11 in macOS in build an source versions: never happened. So I don't see anything that I can do about it.

joltcoke:
The problem doesn’t occur with the source version, only with the build version. Sorry, forgot to mention that.

Does that indicate a problem with my build?

peuter:
[...]

PR #686, opened 2017-12-23
call first makePageSizeValid immediately instead of queueing it

### LINKED ISSUE #663: Diagram popup
Two issues:
In a touch environment a singe touch should scroll the page and not the (big) diagram. (Unless it's in the full page pop up)

The popup of the (big) diagram shows an empty diagram frame on a Galaxy S7. The other diagram popups (inline and button) are working fine as well as all popups on a tablet (older Android)

peuter:
closes #663, #681

### DIFF
--- a/source/class/cv/ui/layout/ResizeHandler.js
+++ b/source/class/cv/ui/layout/ResizeHandler.js
@@ -46,6 +46,7 @@ qx.Class.define('cv.ui.layout.ResizeHandler', {
     $navbarBottom: null,
     width: 0,
     height: 0,
+    __initial: true,
 
     validationQueue: [],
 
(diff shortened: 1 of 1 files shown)


ChristianMayer:
I can't test it right now, so just a question: is it made sure that this
resize is called after(!) the DOM was created and the CSS has been loaded?

Am 23.12.2017 2:47 nachm. schrieb "Tobias Bräutigam" <
notifications@github.com>:

peuter:
Yes the initial makePagesizeValid is (indirectly) triggered by the first scrollToPage (which opens the start page) and that one is called after DOM is ready and everything is loaded.
Queuing the makePagesizeValid leads to the execution of the diagram initialization before calculating the correct page size and thats where the errror occurs, because the diagram DOM element has a width of 0 until the page size is correctly set, and flot doesn't seem to like that.

PR #691, opened 2018-03-15
defer diagram initialization to make sure that the sizes are calculated

peuter:
refs #681

### DIFF
--- a/source/class/cv/plugins/diagram/Diagram.js
+++ b/source/class/cv/plugins/diagram/Diagram.js
@@ -126,9 +126,7 @@ qx.Class.define('cv.plugins.diagram.Diagram', {
               if (!this._init) {
                 this.loadDiagramData(this.plot, false, false);
               } else {
-                qx.bom.AnimationFrame.request(function () {
-                  this.initDiagram(false);
-                }, this);
+                this.initDiagram(false);
               }
               this.removeListenerById(this.__vlid1);
               this.__vlid1 = null;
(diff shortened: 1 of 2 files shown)


ChristianMayer:
It it guaranteed that delaying by one step is allways sufficient or should the delay wait for some event?

peuter:
Well I can't give a guarantee, because the whole size calculation process is quite complex. All I can say, that when the error occured the "makePageSizeValid"-process (which has to be finished before the diagram gets initialized, otherwise the describe error occurs), has been started but not finished when the error occured. So delaying the diagram inititialization by one frame did relyably fix the problem.

When there are other ways to run into the error, we can have a look if we can find a more general fix. But as I am not able to reproduce the problem with the given fix, I think we can use this for now.

ChristianMayer:
So this points in the direction that I had in mind :)

Wouldn't it be a better solution to make this init wait for makePageSizeValid being completed? I.e. make makePageSizeValid send an event when it finishes and make the diagram init delay till this event occurs.

In the past I had cases where a little delay helped - but only till some special case made it fail again. So when we know the real dependency we should use that information.

peuter:
Well if you can tell me where/when this event should be fired and can add an event. I have no idea where to add that. CometVisu's ResizeHandler is finished with its makePageSizeValid, so firing an event there would be too early and does not fix the problem. So my initial error analysis was wrong.

It might be a problem in the flot library, but I am just guessing here.

I would prefer to use this simple fix instead of debugging an unmaintained library which we intent to replace in the long run.

peuter:
How do we proceed with this now? As written, this is still my currently preferred solution to solve the bug. The future proof way would be to get rid of flot and replace it with d3 to create the charts. But I don't have the time to do that in the near future.

PR #710, opened 2018-08-04
[ci nightly] update qx ref to get the fix for:

peuter:
https://github.com/qooxdoo/qooxdoo/issues/9481

reported here
https://knx-user-forum.de/forum/supportforen/cometvisu/1252409-aktuelle-cv-version?p=1253009#post1253009

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,6 @@ before_install:
 install:
 - npm install -g grunt-cli
 - npm install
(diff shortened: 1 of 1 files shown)


peuter:
There isn't much to review here as its just a submodule update.

peuter:
Ok it seems that some things need to be changes in our travis script.

PR #712, opened 2018-08-12
handle missing i values in long-polling transport

### LINKED ISSUE #709: Handle missing "l"
We should check and handle is a missing "l", i.e. 404, as that is just stopping the CV. 

See also #706

peuter:
closes #709 

Please test this code: Because I only can test with openHAB-Backend and that does not use long-polling but sse and sse does not need the "i"-value. So => I wasn't able to test this code.


client/source/class/cv/io/Client.js
```diff
@@ -92,7 +92,9 @@ qx.Class.define('cv.io.Client', {
     TEST_MODE: false,
     ERROR_CODES: {
       CONNECTION: 1,
-      PROTOCOL_MISSING_VERSION: 10
+      PROTOCOL_MISSING_VERSION: 10,
+      PROTOCOL_INVALID_READ_RESPONSE: 50,
+      PROTOCOL_INVALID_READ_RESPONSE_MISSING_I: 51,
```
ChristianMayer: The automated build doesn't run - I guess do to this trailing comma

PR #721, opened 2018-08-20
shorter stack traces:

### LINKED ISSUE #720: Stack trace report too long
When an error with a long stack trace happens and the user clicks on the report button the recieving side at github gives an `414 Request-URI Too Large` error.

peuter:
- do not print duplicate lines
- cut-off trace if it exceed a certain limit

closes #720


peuter:
Ok, done.

PR #726, opened 2018-09-19
add include files to config_check by resolving the includes …

peuter:
…to build a complete config file and check that one.

Errors that are found in the included files will be reported with included file name and the line number in that file.

### DIFF
--- a/source/check_config.php
+++ b/source/check_config.php
@@ -33,9 +33,11 @@ function xml_highlight($s)
 
 function libxml_display_error( $error )
 {
-  global $lines, $error_array;
+  global $combinedLines, $error_array, $lineMappings;
 
   $error_array[] = $error->line;
+
+  $return = "";
 
   switch ($error->level)
   {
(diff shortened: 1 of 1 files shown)


source/check_config.php
```diff
@@ -133,10 +142,47 @@ function checkVersion( $dom )
...
 $lines = file( $conffile );
-$dom->load( $conffile );
+$combinedLines = $lines;
+$lineMappings = array();
+$mappedLineOffset = 0;
+
+foreach( $lines as $line_num => $line ) {
+    if (preg_match('/(\s*)(.*)<include src="([^"]+)"\s*\/>(.*)/', $line, $matches)) {
+        $includedLines = file($resourceDir . $matches[3]);
+        $func = function($value) {
```
ChristianMayer: Anonymous functions were introduced with PHP 5.3 - but the WireGate runs PHP Version 5.2.6-1+lenny16
=> Please change to ordinary function

peuter:
Ok both problems should be fixed now

PR #729, opened 2018-09-23
Add sentry error reporting feature

peuter:
to enable sentry-support the user has to load the index-sentry.html file. All unhandled errors will be (automatically) reported to senty then. Captures error will show a button to send the error to sentry.


source/class/cv/Application.js
```diff
@@ -307,9 +307,27 @@ qx.Class.define("cv.Application",
...
+            needsConfirmation: false
+          }
+        );
+      }
       cv.core.notifications.Router.dispatchMessage(notification.topic, notification);
     },
 
+    throwError: qx.core.Environment.select('qx.globalErrorHandling', {
+      "true":  function () {
+        window.onerror(new Error('test error'));
```
ChristianMayer: Is "test error" intended or a left over?

peuter: No this one is intented to have an easy way to test the globalErrorHandling (when its enabled)

ChristianMayer:
Can you describe what I should do to test this / make it work?

peuter:
Well the normal User doesn`t see whats happening in the Background. You need a sentry account and the I have to add you to the CometVisu-project on sentry. Then you get an Email about the error and can see on overview on all errors.

I got the mail from the error you provoced ;-)

peuter:
`index-sentry.html` has been replaced by `reportErrors` URL parameter now.

ChristianMayer:
grafik

grafik

- The report to GitHub should be removed now
- The two lines should be out outside the scrolled area as the user won't find it by himself otherwise (I don't expect that a user is scrolling here as it is gibberish to him)
- The second line should be translated as well (or the first shouldn't... just make it consistent)

peuter:
All addressed problem should be solved now

PR #739, opened 2018-10-24
add dockerfile and nightly builds of docker images

peuter:
- this PR introduces build stages: lint, test and deploy
- a deploy job for creating and pushing a docker image for nightly builds have been added.
- as the unit tests are a but picky some changes for them are included too, but is still can happen that the test fails, needs further investigation, but this PR improves the situation already

Whats is needed to make the docker builds/pushes work is a dedicated docker-hub account that is allowed to push to cometvisu/cometvisu. The credentials of this account need to be added as encrypted envronment variables to travis, like this (travis cli client is needed):

```
travis env set DOCKER_USER myusername
travis env set DOCKER_PASS secretsecret
```

This command line need to be executes in the local clone of the cometvisu-repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 language: python
 dist: trusty
 services:
-  - docker
+- docker
 python:
 - '3.5'
 cache:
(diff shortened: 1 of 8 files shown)


peuter:
Everything is prepared to make this work. Ready to be merged.

PR #743, opened 2018-10-27
implement an overlay filesystem in resource/config volume that allows…

### LINKED ISSUE #738: Move user changeable files in one directory
The user changeable files like configs but also e.g. custom.css should all be located in one directory to make it easy to migrate to a new version, create a Docker container where those files are located in a Volume, ...

This would also solve: https://forum.timberwolf.io/viewtopic.php?p=2409#p2409

peuter:
… all files in the resources subfolder to be overridden.

Uses rewrite rules to check if a file like resource/designs/metal/custom.css exists as resource/config/overlay/designs/metal/custom.css exists and delivers this one. The original file will be delivered as fallback if it does not exists as overlay file.

closes #738

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,9 +107,11 @@ RUN { \
     } | tee "/usr/lib/cgi-bin/l" \
  && chmod +x /usr/lib/cgi-bin/l \
  && a2enmod cgi \
- && a2enmod headers
+ && a2enmod headers \
+ && a2enmod rewrite
 
 COPY build/ /var/www/html/
+COPY utils/docker/_.htaccess /var/www/html/.htaccess
 
 # Options - especially for development.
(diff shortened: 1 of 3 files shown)


ChristianMayer:
Instead of an "complicated" overlay system I'd prefer to put all customer modifiable content in one  place.
If e.g. the content would be located in the media directory the existing infrastructure (manager.php) could be used to manipulate it.

So my suggestion is rather to extend the config file to have an (optional) attribute for a name of a custom.css and an attribute for a custom.js where both are looked for in the config/media/ directory. 

This would also allow for different custom files for different configs at the same time. And everything where we expect the user to do some modification can be done by the manager.php and no direct file access is needed.

Any more advanced modifications would require direct access but that could then be expected from the user.

PR #744, opened 2018-10-27
Enhance docker labels

ChristianMayer:


utils/travis/docker_push.sh
```diff
@@ -18,7 +18,10 @@ fi
 
 echo "building docker container for ${IMAGE_NAME}:${VERSION_TAG},${MASTER_TAG},${SUB_TAG} ..."
 
-docker build -t $IMAGE_NAME:$VERSION_TAG .
+BUILD_DATE=`date --iso-8601=seconds`
+VCS_REF=${TRAVIS_PULL_REQUEST_SHA}
+
+docker build -t $IMAGE_NAME:$VERSION_TAG --build-arg BUILD_DATE --build-arg VERSION_TAG --build-arg TRAVIS_JOB_NUMBER --build-arg TRAVIS_PULL_REQUEST.
```
peuter: Does `TRAVIS_PULL_REQUEST` make sense here. We do not build containers from pull requests and therefore `TRAVIS_PULL_REQUEST` will always be `false`, as described here: https://docs.travis-ci.com/user/environment-variables#convenience-variables.

`TRAVIS_PULL_REQUEST_SHA`would also be emtpy, I guess `TRAVIS_COMMIT`is what you are looking for (but I am not sure, needs to be tested).

PR #748, opened 2018-10-30
do not send values that are received via update (and debug written da…

### LINKED ISSUE #747: Make widgets send only after user action
According to https://knx-user-forum.de/forum/supportforen/cometvisu/1252409-aktuelle-cv-version?p=1280635#post1280635 it seems that the color chooser and/or slider will also send based on an update from the bus.

Widgets should only send based on user action.

peuter:
…te in testMode for easier debugging)

closes #747


peuter:
Should be fixed now-

PR #779, opened 2018-12-14
Enhance config https

ChristianMayer:


source/resource/demo/visu_config_gauge.xml
```diff
@@ -132,7 +132,7 @@
           <layout colspan="1.3"/>
           <address transform="DPT:16" mode="read" variant="setValue">12/7/95</address>
         </gauge>
-        <switch mapping="Text_Scroll" styling="- not set - (undefined)" on_value="CometVisu Geil" off_value="AUS" bind_click_to_widget="true">
+        <switch mapping="Text_Scroll" on_value="CometVisu Geil" off_value="AUS" bind_click_to_widget="true">
```
peuter: What does the styling setting mean, is this on purpose?

ChristianMayer: Nothing. It's a left over when the Editor is used "wrongly". (Actually it's an editor bug in my point of view)

peuter: Shouldn't we remove that from the demo + example configs then?

ChristianMayer: Yes - and this PR is doing that. Or did I miss any?

peuter: You're right. Seems that am I not able to read pull requests anymore ;-)

PR #791, opened 2018-12-23
add templating feature for config files

peuter:


ChristianMayer:
- Travis CI build failed
-  `source/resource/config/heating.template.xml` should be located at `source/resource/demo/` when it's a demo. Currently it's not used.
- There seems to be no library (mustache) getting included. I guess that's missing in the PR.
- Before using an external library please consider whether we can achieve nearly the full functionality (or even the full functionality of what we need) by implementing it our self. - The (release build) code should be so small that (mobile!) browsers like to keep it in the cache to prevent any loading over a slow mobile network (1 bar EDGE...) and only keep the volatile data being transmitted.

peuter:
The mustache code is embedded in `qx.bom.Template` which is basically just a wrapper for mustache. This  PR also does not add anything, because the templates are already used in the NotificationCenter.

And I prefer to use a library that is used and tested by many people than writing everything our own. The mustache-Code adds about 9.6kb (minified). I dont think its worth the effort just to save some data. I would agree with you, when its about using libraries with >100kb but not in this case.

Even if we save some Kb by writing something our own which provides only the things we need here. I assume that in the long run we will end with about the same features as mustache provides, because once users will start using the templates they will demand more features. And either we deny those feature requests or we end in the same amount of code (or even more) than we have right now using mustache.

ChristianMayer:
The file was moved now: `source/resource/config/heating.template.xml` → `...esource/config/media/heating.template.xml`
Bus shouldn't it be located in `source/resource/demo/` as it's a demo?

peuter:
Sorry, I removed the file, because it isn't used in this PR at all. I use it in my current experiment with the Config-File for the Demo-Page. So it will be part of that PR, when  its ready.

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
...
+                      <icon name="temp_control" />
+                      Soll
+                    </label>
+                    <address transform="OH:number" variant="">{{ targettemp_address }}</address>
+                  </infotrigger>
+                </group>
+            </template>
+        </meta>
+        <pages...>
+            <page name=="Wohnzimmer"...>
```
ChristianMayer: `name=="Wohnzimmer"` -> `name="Wohnzimmer"`

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
...
+                ...
+                <template name="Heizung">
+                  <value name="control_address">Heating_FF_Living</value>
+                  <value name="currenttemp_address">Temperature_FF_Living</value>
+                  <value name="targettemp_address">Temperature_FF_Living_Target</value>
+                </template>
+                ...
+            </page>
+            <page name=="Küche"...>
```
ChristianMayer: `==` -> `=`

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
...
+.. code-block:: xml
+    :caption: Beispiel einer Template-Definition aus einer externen Datei
+
+
+    <pages...>
+        <meta>
+            <template name="Heizung" ref="resource/config/media/heizung.template.xml"/>
+        </meta>
+        <pages...>
+            <page name=="Wohnzimmer"...>
```
ChristianMayer: `==` -> `=`

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
         </statusbar>
     </meta>
 
+.. _xml-format_templates:
+
+Templates
+---------
+
+Im Metabereich können Templates für oft verwendete Konfigurationsausschnitte erstellt werden. In der Regel möchte z.B.
```
ChristianMayer: `Regel möchte z.B.` -> `Regel möchte man z.B.`

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
...
+seine Heizungs in jeden Raum auf die gleiche Weise darstellen. Diese kann aber aus mehrere Widgets bestehen, z.B: einem
```
ChristianMayer: `z.B:` -> `z.B.`

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
...
+            <page name=="Wohnzimmer"...>
+                ...
+                <template name="Heizung">
+                  <value name="control_address">Heating_FF_Living</value>
+                  <value name="currenttemp_address">Temperature_FF_Living</value>
+                  <value name="targettemp_address">Temperature_FF_Living_Target</value>
+                </template>
+                ...
+            </page>
+            <page name=="Küche"...>
```
ChristianMayer: `==` -> `=`

doc/manual/de/config/xml-format.rst
```diff
@@ -168,6 +168,149 @@ Der Statusbar befindet sich am unteren Bildschirmrand und erlaubt das zB. Abzeig
...
+        <meta>
+            <template name="Heizung">
+                <group name="Heizung">
+                  {{{ additional_content }}}
+                  <slide min="0" max="100" format="%d%%">
+                    <label>
+                      <icon name="sani_heating" />
+                      Heizung
+                    </label>
+                    <address transform="OH:dimmer" variant="">{{ control_address }}</address>
```
ChristianMayer: The XSD doesn't accept such an `<address>` content

PR #802, opened 2019-01-02
do not increase blocker counter for unique messages: This unblocks th…

### LINKED ISSUE #795: Remove screen block when error got resolved
When an error occurs (e.g. failed "r" with a 504) the notification center shows correctly the error.
But when the error gets resolved by itself (e.g. a new "r" works fine) the grey screen block is still shown.

During error:
grafik

Post error:
grafik

peuter:
…e UI after the message disappears no matter hpow often it got blocked before.

 closes #795

### DIFF
--- a/source/class/cv/core/notifications/Router.js
+++ b/source/class/cv/core/notifications/Router.js
@@ -189,9 +189,10 @@ qx.Class.define("cv.core.notifications.Router", {
      * @param address {String} GA or openHAB item name
      * @param state {var} received State
      * @param initial {Boolean} true id this is the first state update for this address
+     * @param changed {Boolean} true if the incoming state update differs from the last one
      * @protected
      */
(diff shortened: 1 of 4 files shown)


ChristianMayer:
Could it happen that we have unique as well as non unique messages at the same time?
Especially could it happen that a few non unique messages are existing when a unique comes and gets resolved and thus removes the block for the still existing ones?

peuter:
Yes you are right, that situation can happen. I' going to change this PR to handle that.

peuter:
Ok next try: Theoretically it is still possible to have unique-messages unblocking non-unique ones with the same topic. But because popup messages for the same topic are either all unique or not, this will never be a problem pratically.

Please test this with the real connections errors as I couldn't check if the topic gets stored in the DO Element to have it avaible for the unblock-Call.

Issue #625, opened 2017-10-04
Plugin compability with 0.11.x


peuter:
This issue should track the compability state of the plugins:

- [x] Diagram
- [x] openHAB
- [x] CalendarList
- [x] Clock
- [x] ColorChooser
- [x] Gauge
- [x] Link
- [x] MobileMenu
- [x] Openweathermap
- [x] PowerSpectrum 
- [x] Rss
- [x] RssLog
- [x] Speech
- [x] Strftime
- [x] Svg
- [x] Timeout
- [x] UpnpController

ChristianMayer:
- **CalendarList** is failing with `Uncaught Error: Error in property days of class cv.plugins.CalendarList in method $$setDaysImpl with incoming value 'null': Null value is not allowed!` (testing with dummy data; but even then it mustn't fail like that)
  Docu might be: https://knx-user-forum.de/forum/supportforen/cometvisu/33952-google-kalender-in-der-cometvisu
  Criticality is **high** as I think it is used in the field
- **Clock** is failing during page load at line https://github.com/CometVisu/CometVisu/blob/develop/source/class/cv/plugins/Clock.js#L45
  `Uncaught TypeError: this.getAttributeToPropertyMappings is not a function`
  Criticality is **low** as I think it is not used in the field (plugin was mostly a proof of concept)
- **MobileMenu** is failing with `MobileMenu.js:48 Uncaught TypeError: this.getAttributeToPropertyMappings is not a function`
  Docu might be: https://knx-user-forum.de/forum/supportforen/cometvisu/880976-beta-material-design
  Criticality is **unknown** as I have no idea how it is used in the field (probably **middle**)
- **Openweathermap** is failing with `Uncaught Error: Error in property cssClass of class cv.plugins.OpenweatherMap in method $$setCssClassImpl with incoming value 'null': Null value is not allowed!`
  I couldn't find any docu
  Criticality is **unknown** as I have no idea how it is used in the field (probably low or middle)
- **UpnpController** is failing with `Uncaught Error: Error in property playerIp of class cv.plugins.UpnpController in method $$setPlayerIpImpl with incoming value 'null': Null value is not allowed!`
  Docu(?): https://knx-user-forum.de/forum/supportforen/cometvisu/18386-sonos-contoller-als-cv-plugin-oder-allg-upnp-contoller
    Criticality is **unknown** as I have no idea how it is used in the field (probably low or middle)

ChristianMayer:
[...]

PR #809, opened 2019-01-06
Fix plugins

peuter:
refs #625


peuter:
Ok, now it works even if no refresh value is defined.

PR #818, opened 2019-01-13
Fix more doc problems and be more verbose during generating

ChristianMayer:


utils/commands/doc.py
```diff
@@ -195,6 +198,9 @@ def _run(self, language, target_dir, browser, skip_screenshots=True, force=False
                   "--target=%s" % screenshot_build, _out=self.process_output, _err=self.process_output)
 
             # 2dn run with access to the generated screenshots
+            print '================================================================================'
+            print 'sphinx_build: second run'
+            print '================================================================================'
             sphinx_build("-b", target_type, source_dir, target_dir, _out=self.process_output, _err=self.process_output)
```
peuter: These print statements are not python3 compatible, please use `print('..')` which works in both python versions.

utils/commands/doc.py
```diff
@@ -186,6 +186,9 @@ def _run(self, language, target_dir, browser, skip_screenshots=True, force=False
             os.makedirs(target_dir)
 
         # first run generates the widget-example configs
+        print '================================================================================'
+        print 'sphinx_build: first run'
+        print '================================================================================'
```
peuter: Also => `print('..')`

PR #825, opened 2019-01-18
make sure that width: auto gets applied in every mode

peuter:


ChristianMayer:
What is the reason for this PR? What failure does it fix?

peuter:
I thought it would be a bugfix which is relevant in the current code, but actually it just fixes a problem in a feature that is currently only used in the currently developed demo-mode. So there is no need for a seperate PR => closing it. Will be part of the demo-mode PR, whenever it will be ready.

PR #826, opened 2019-01-18
Fix notification docu

peuter:


peuter:
Oops, I have removed that change from the PR.

Issue #817, opened 2019-01-13
Demo mode enhancements


ChristianMayer:
Out of the review of #794: 
- [x] Temperatures should be different between the rooms
- [x] All rooms should exists (Büro/Kinder/Flur)
- [x] Slider doesn't slide by mouse (Chrome), similar to the tablet issue
- [x] pressed/unpressed button state looks too similar. The difference isn't easily visible
- [x] Wohnen: Licht vs. Licht mit Icons - it would be better to show what's possible in a "show off demo" than to educate about subtle differences
- [x] Refresh doesn't make sense in such a demo
- [x] Make demo more interactive, i.e. let switches switch something somewhere else to give better feedback (e.g. ColorChooser -> r+g+b slider; main light slider -> measured brightness in Lux; ...)

Mode possibilities (requiring deeper code changes):
- [x] Add time dimension (+ infinite loop) to testmode data json so that something is happening
- [x] Add "callback functions" to testmode data which can react to sent data by sending something else back. This could allow an interactive, simulated dim or shade. Even complex behavior like room temperature could be simulated then

peuter:
I will open an extra issue for task 3, because I think it is a bug that does not only affect the demo mode (but I need to check that first). Task 4 is a design decision, which is also not part of the demo mode, thats how the metal design visualizes buttons since years I think and I'm fine with that.

All other tasks should be somehow handled in the new PR.

ChristianMayer:
It's solved

PR #830, opened 2019-01-19
Demo mode

peuter:
this should implement most of the tasks listed in #817.

Also includes 2 fixes for the gauge plugin: The address variant values are parsed now and there was an issue with the initualization order wich could cause lost update values, because handleUpdate method weas called before the gouge element was available.

Also introduces an responsive boolean `scale` attribute for layouts to prevent element positioning on 2d pages from beeing dependent from the bacdrop scale value. Is used on the start page for the three groups on the right side.

There is a basic time based sequence in the test data that is player in an infinite loop, to show some "update-action". This is just a start and can be enhanced by adding more steps.

And there is a simple device simulation (currently only for shutters), which send shutter positions in a sequence when you drive the shutter up or down, which is visualized by a slider in the visu. Can be tested with the two "Jalousien" in "Wohnzimmer".


source/resource/demo/media/demo_2d_backdrop_red_pot.js
```diff
@@ -10,9 +10,13 @@ var
 client.update = function(json ) // overload the handler
 {
   var h = cv.Transform.decode(thisTransform, json[thisGA] );
+  if (h === undefined) {
+    return;
+  }
   var filling = qx.bom.Selector.query('#rect3855', document)[0];
   filling.y.baseVal.value=200.57388 + (100-h)*2;
   filling.height.baseVal.value = h*2;
+  console.log(h);
```
ChristianMayer: Leftover that should be removed?

PR #857, opened 2019-02-07
add option to send the long value before mouse release (direcly when …

peuter:
…the threshold is exceeded)

### DIFF
--- a/source/resource/visu_config.xsd
+++ b/source/resource/visu_config.xsd
@@ -384,6 +384,13 @@
       <xsd:documentation xml:lang="de">Beim Aktivieren dieser Option wird die gesamte Widget Fläche als Schaltfläche genutzt</xsd:documentation>
     </xsd:annotation>
   </xsd:attribute>
+
+  <xsd:attribute name="send-long-on-release" type="xsd:boolean">
+    <xsd:annotation>
+      <xsd:documentation xml:lang="en">Fire the long press event after the button is released (=true) or immediately after the threshold is exceeded (=false).</xsd:documentation>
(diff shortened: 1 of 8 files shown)


source/resource/visu_config.xsd
```diff
@@ -384,6 +384,13 @@
       <xsd:documentation xml:lang="de">Beim Aktivieren dieser Option wird die gesamte Widget Fläche als Schaltfläche genutzt</xsd:documentation>
     </xsd:annotation>
   </xsd:attribute>
+
+  <xsd:attribute name="send-long-on-release" type="xsd:boolean">
+    <xsd:annotation>
+      <xsd:documentation xml:lang="en">Fire the long press event after the button is released (=true) or immediately after the threshold is exceeded (=false).</xsd:documentation>
+      <xsd:documentation xml:lang="de">Senden den Wert für den langen Tastendruck erst nachdem der Button losgelassen wurde (=true), oder sofort nachdem der Schwellwert für den kurzen Tastendruck überschritten wurde (=false).</xsd:documentation>
```
ChristianMayer: "Senden" -> "Sendet"

peuter:
I'm not sure that I correctly understand what you mean with "breaks the longtime-Sending". The only bug I've found is that a long-click on the down-button did not send the downvalue, but the upvalue instead. That is fixed now. Despite that problem I did not see something wrong.

ChristianMayer:
That fixed it :+1:

peuter:
Hm I dont have those errors, when I run the tests locally. Kind of hard to find the problem this way.

ChristianMayer:
A quick check on my system also went smoothly. But with a failing Travis I can't commit :(

peuter:
Now Travis is happy, this was a really tricky one ;-)

PR #879, opened 2019-02-28
no undefined checks because the values might be null, fixes error whe…

peuter:
…n clicking on "Temperatur" in demo_flat.


peuter:
> It would be cleaner when the code does an explicit check for what's needed (a string) and not depend on implicit type conversion

Done

Issue #894, opened 2019-03-03
Handle refreshs nicer


ChristianMayer:
For the diagram plugin: shift display with data update
-> https://knx-user-forum.de/forum/supportforen/cometvisu/1334235-diagrammaktualisierung-funktioniert-nicht

For the tr064 plugin: sync refresh with active page and take care when the SID might expire
-> https://knx-user-forum.de/forum/supportforen/cometvisu/1315369-fritzbox-plugin

PR #895, opened 2019-03-05
improve refresh stop/start handling when widget changes its visibilit…

peuter:
…y state. Start timer immediately if the widget gets visible and the last runtime is older then the refresh interval, otherwise start when the next interval step is reached.

Removes the custom refresh start/stop handling if the diagram widget to avoid duplication behaviours.

refs #894


peuter:
conflict solved

PR #896, opened 2019-03-06
Fix refresh behaviour

peuter:
- Diagram refresh has been fixed: the topic was wrong, so the AbstractDiagram._startRefresh method has never been called
- do not let the Refresh-Mixin start/stop timers that have been created somewhere else. This should prevent the refresh function from stopping in tr064 (and other) plugin.

Edit:
- just changed the behaviour to refresh immediately, after the diagram page appears

### DIFF
--- a/source/class/cv/ui/common/Refresh.js
+++ b/source/class/cv/ui/common/Refresh.js
@@ -37,7 +37,6 @@ qx.Mixin.define("cv.ui.common.Refresh", {
     // Stop the while invisible
     this.addListener("changeVisible", function(ev) {
       if (this._timer && ev.getData() !== ev.getOldData() && this.__ownTimerId === this._timer.toHashCode()) {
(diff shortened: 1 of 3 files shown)


source/class/cv/ui/common/Refresh.js
```diff
@@ -36,7 +36,8 @@ qx.Mixin.define("cv.ui.common.Refresh", {
 
     // Stop the while invisible
     this.addListener("changeVisible", function(ev) {
-      if (this._timer && ev.getData() !== ev.getOldData()) {
+      if (this._timer && ev.getData() !== ev.getOldData() && this.__ownTimerId === this._timer.toHashCode()) {
+        console.log('timer', this.getPath());
```
ChristianMayer: Left over debug output?

peuter: removed

Issue #926, opened 2019-04-14
Improve documentation structure


peuter:
- [x] change current `current` to `latest`
- [x] build "version"-doc from develop branch and add a symlink named `develop`to the generated version (formerly known as `latest`)
- [x] Add canonical tags to the pages using the version number as link (to avoid duplicates in `latest` and `develop`
- [x] Create sitemap for docu

PR #933, opened 2019-04-19
Change docu symlinks

ChristianMayer:
See also #926 action 1 and 2


utils/commands/doc.py
```diff
@@ -215,6 +215,39 @@ def _run(self, language, target_dir, browser, skip_screenshots=True, force=False
...
+            # newer version.
+            # Simple work around: just retrigger a build of the newer version
+            # afterwards manually.
+            print('detected build of most recent version')
+            symlinkname = self.config.get("DEFAULT", "most-recent-version-mapping")
+
+        if '' != symlinkname:
+            print("setting symlink '%s' to '%s'" % (symlinkname, target_dir))
+            cwd = os.getcwd()
+            os.chdir(root)
```
peuter: This line throws an error when i generate the documentation:
`NameError: global name 'root' is not defined`

ChristianMayer: Fixed in 9d65b21

Issue #782, opened 2018-12-16
Editor 2.0


ChristianMayer:
The bring the editor to the next level a few points should be done:

- [ ] _internal:_ Make code `strict`
- [ ] _internal:_ Separate view and logic, e.g. use a tree widget
- [ ] _visible:_ Allow drag&drop to move elements around (instead of the current `sort`)
- [ ] _visible:_ make it more mobile device friendly (-> no context menu to create new elements)
- [ ] _visible:_ Link editor with preview to that the use immediately sees where the selected element is in the visu
- [ ] _visible:_ replace all alert boxes with less disturbing/distracting feedback
- [ ] _visible:_ Allow language switch and implement mulitlanguage

peuter:
I guess this is done with the new xml editor

Issue #722, opened 2018-08-24
Refactor manager.php


ChristianMayer:
It has grown too much (especially with the TR-064 plugin)

peuter:
I am currently working on a replacement for all the php stuff (thats not part of a plugin). The main purpose it to reduce the code that needs to run on the server as much as possible. The "new" Backend-Code just provides a REST-API with basic CRUD operations for files/folders in the config-folder and support for the hidden configuration modifications.
The new manager is something like a Web-IDE, where you can code your CSS/JS/Config-XML files, having all the fancy stuff like code-completion, validation. Quite nice already at this early stage of development. Still a long way to go but worth the effort I think.
The current XML-Editor is currently embedded as an IFrame and feels like an alien in there. I guess it needs some love too in the near future as mentioned in #782.

ChristianMayer:
As already written somewhere else, we should have a sort of CI for the CV "support" pages. Currently it's very cluttered: one design for the homepage, one for the manual, one for the API (ok, users don't go there), one for the manager, one for the graphical editor, one for the text editor...

Personally I don't have a CI yet, except perhaps the homepage (which wasn't really desinged, it just "happened")

peuter:
Done with the new manager

PR #947, opened 2019-06-25
Add REST-API and a new manager that uses it

peuter:
replace the old manager by splitting its features into a backend part provided by a REST-API (which is defined with an openapi/swagger file). And a new Javascript based client which uses this API to provide the same features (and more) as the old PHP-based manager + the two Editors.

The basic goal was to keep the backend as simple as possible and implement most of the logic in the client (because the backend needs to be implemented twice once for the "normal" use and once for the openHAB backend).

**Backend:**
- Currently implemented with node. The initial idea was to get a replacement for PHP, which also work in the JAVA-based openHAB backend. It turned out later that this is hard to achieve and it will be easier to re-implement the backend in Java. A re-implementation in PHP for the other backends might by an option, but I didn't like the PHP support for swagger (basically needs symfony framework to work). This is a point to be discussed.
- replaces the dataproviders implemented in PHP

**Frontend:**
- Replaces the old manager in Javascript, does not need to reload the page to start the manager. Implemented with Qooxdoo's UI tools.
- Tight integration of the XML-Text-Editor
- Basic integration of the old XML-Editor (uses the dataproviders and saving from the REST-API) => a task for the future would be to re-implement the XML-Editor for better integration (see #782).
- Icon overview is included
- You can compare a config file with one of its backups to see the changes
- Added a trash feature (files are not deleted directly but moved to the trash which can be emptied)
- There are lots of other new features, just try it out.

The new manager can be started either by pressing `Ctrl+M` (this keycode also closes the manager) in a loaded CometVisu or by adding `manager=1` to the Querystring.

# Getting started
1. `git submodule init` + `git submodule update` because the are new submodules
2. `npm install` new /updated depedencies
3. `./generate source'
4. Start backend with `node source/rest/index.js`
5. Load cometvisu with `manager=1` in the query string.
6. Play around with the manager

This PR is a first shot for this quite complex change. Its basically to find missing features and critical errors and get a basis for further development.

refs #722


ChristianMayer:
npm is missing express, cors and exegesis-express to be able to run the server

ChristianMayer:
- Scrolling doesn't work - so part of the manager page isn't visible
- Showing "file icons" is wasting lots of space and is uncomfortable to find the right file. A simple list would be much better
- Sometime errors appear on the console
- Generally I don't like the use of even more Qx, which makes the CV even fatter - we should aim to get smaller!
- Introducing npm on the server makes the server fatter and harder to install. That should only be an additional/different way to run the server and not be a requirement to run the CV

As the introduction of Qx to the CV made it to an not maintainable mess to me I strictly oppose the introduction of even more Qx.

peuter:
> * Scrolling doesn't work - so part of the manager page isn't visible

I will have a look at this. May take some time.

> * Showing "file icons" is wasting lots of space and is uncomfortable to find the right file. A simple list would be much better

An alternate View (as default) as list shouldn't be a problem

> * Sometime errors appear on the console

I will have a look at this, Maybe I need more detailed information later.

> * Generally I don't like the use of even more Qx, which makes the CV even fatter - we should aim to get smaller!

As the Manager-Code is only loaded on demand (on first use) the "normal" CV isn't affected at all. You cannot add more functionality with less code.

> * Introducing npm on the server makes the server fatter and harder to install. That should only be an additional/different way to run the server and not be a requirement to run the CV

As I said the server part is a point to be discussed. But I don't agree with the point that its harder to install. In a release all you have to do (when its implemented that way) to call `npm start`. The user does not have to install an extra server with PHP support. So that is even easier but the downside is that existing installations need to change something to get the new manager support (when they don't use docker). As I said, this is a point to be discussed. If we come to the conclusion that we want to stick to PHP I'm fine with that. We will have at least two REST-server versions (the openHAB one in Java and the PHP or node one for the others), thats the reason why I tried to keep the REST-server as easy as possible.

> As the introduction of Qx to the CV made it to an not maintainable mess to me I strictly oppose the introduction of even more Qx.

Introducing a framework and then avoid using it doesn't make sense at all to me. If I would have wrote it all by myself you would have the same amount of code with less/worse documentation and way more errors. And you would have to learn/understand the code anyways to maintain it. 
The main advantage of a framework is to have useable, battle tested code that helps you to get things done in a reasonable amount of time.

ChristianMayer:
I think we should split what we want to achieve from the realization first:

- The Java based server environment as used by OpenHAB has troubles with providing PHP support: ok, understood
- Splitting up the manager between server side and client side by e.g. a REST protocol creates a level of abstraction that could help here: yes, makes sense
- 1. - Refactor the manager on the client side is necessary as the current code was just a quick fix that kept growing 
      - How to do it and which toolkit to use (when necessary) is still open to discussion. It could be that Qx is the best solution, but that's not god given. It really depends on the desired complexity and what is e.g. possible with native infrastructure. For the current functionality I tend to say that native should still be very simple.
  1. - The server side (the REST server) should take different environments into account
      - So I would suggest to have the server functionality so simple that different implementations could live parallel together. One in PHP for best backward compatibility and one that is best for OpenHAB use (e.g. `npm`)

The problem with Qx that I see is that:
- I can't maintain the code anymore as it creates failure that I don't know how to fix. And I'm trying hard.
- it is limiting us instead of enhancing our possibilities as it is preventing the use of modern ECMAScript
- it is duplicating stuff that the native language is (now) already providing, e.g. classes. And native language is always better than any lib

peuter:
> I think we should split what we want to achieve from the realization first:

I think we are on the same page which the basic conclusions: A simple REST-Backend that needs to be implemented and maintained twice, one PHP-version which is part of the CometVisu release and one Java Version, which needs to be part of the CometVisu-openHAB binding (I don't think its possible to have this code inside the CometVisu and then somehow dynamically load it in the openHAB backend).

>      * How to do it and which toolkit to use (when necessary) is still open to discussion. It could be that Qx is the best solution, but that's not god given. It really depends on the desired complexity and what is e.g. possible with native infrastructure. For the current functionality I tend to say that native should still be very simple.

Native infrastructure does not provide any UI widgets as Qx (or other UI frameworks) do. I used a lot of them (e.g. different Menus, the tree-widget (you have to open the expert mode to see it on the left side), different list versions, and lots of other helpfull stuff. The funtionality implemented in this PR is already quite complex. If I would have wrote all the widgets and other code is used from Qx by myself, I would have needed at least a year and not 3 months. So please don't underestimate this.

> The problem with Qx that I see is that:
> 
> * I can't maintain the code anymore as it creates failure that I don't know how to fix. And I'm trying hard.

If you are talking about errors in release version (as the ones in sentry for example) I totally agree with you. The stacktraces of the minified code are mostly useless. But in the development (aka source) version I don't agree, usually the Stacktrace helps to find out whats going wrong, if there are asynchronous calls involved its a bit harder, but thats a problem you generally have in javascript. If there is anything I can help you with, please ask, I will do my best to support you.
What we really need are sourcemaps. There is a new Qx-Compiler (https://github.com/qooxdoo/qooxdoo-compiler) which supports that, but as we have some customizations in our code e.g. for the screenshots in the source documentation I have to check how we can migrate that to the new compiler before we could use it. And I didn't had the time to take a closer look at it.

> * it is limiting us instead of enhancing our possibilities as it is preventing the use of modern ECMAScript

Same as above, the new compiler supports modern ECMAScript, even unreleased stuff by using babel as transpiler.

> * it is duplicating stuff that the native language is (now) already providing, e.g. classes. And native language is always better than any lib.

The Qx-Class system provides a lot more features than the native one (Mixins, Interfaces, the property system, ...). If you just compare Qx-Classes to native classes you are right but it is a lot more than that.

So most of the problems you have with Qx are solvable in the future or basically right now, if I only would find the time to do so.

peuter:
Additional note: I would suggest to treat the current node backend as reference implemetation which needs to be replaced by a PHP and a JAVA one ASAP. Once that is done, it can be deleted.

peuter:
- Scrolling problem should be fixed now
- there is a list view mode for the start page, but IMHO the old mode (now called preview) is more clearly as you can see preview images and the file types are easier to distinguish. So I left the preview mode as default. If you change the mode the setting will be stored in the localStorage and will be remembered.

Regading the errors on the console: I haven't seen one, so I need more information about that if they still exists.

peuter:
The node implentation has been replaced with a PHP one. In development version it needs a one time `grunt composer:rest:install` (and an installed composer for php),

ChristianMayer:
This PR is here since quite a while - and the reason is me getting annoyed from the current state of the CV code. I was only seeing the troubles that Qx are causing me and so I was spending my time with stuff that I enjoyed more than working with CV code.

So I see it as my fault that I allowed the CV to switch not just a library - as I thought the jQuery->Qx was aiming at - but also to switch it to a new programming language that Qx actually is (due to it's compiler and it's incompatibility with modern JavaScript). Sorry!
As I don't want to repeat such a failure I will not merge any code that is increasing the Qx use, only code that is getting rid of it. By working on #953 I couldn't see any place where Qx really is helping us. (OK, probably in the OpenHAB plugin where `qx.ui` is used, but a plugin shouldn't take the full CV as a hostage...)

How the CV code should look in the future is OT here, we should discuss it on a different place.

.travis.yml
```diff
@@ -65,9 +65,9 @@ jobs:
   - name: 'End-to-end Tests: Chrome beta'
     env: CV_BUILD=test CV_BROWSER=Chrome_travis CV_VERSION=beta
     script: *2
-  - name: 'End-to-end Tests: Chrome unstable'
-    env: CV_BUILD=test CV_BROWSER=Chrome_travis CV_VERSION=unstable
-    script: *2
+#  - name: 'End-to-end Tests: Chrome unstable'
+#    env: CV_BUILD=test CV_BROWSER=Chrome_travis CV_VERSION=unstable
+#    script: *2
```
ChristianMayer: Does it still has to be removed - or should it stay?

peuter: I don't know, lets give it a try.

peuter: It seems that this is still necessary.

source/resource/config/hidden.php
```diff
@@ -1,6 +1,13 @@
...
 // File for configuraions that shouldn't be shared with the user
 $hidden = array(
+<<<<<<< HEAD
   'fritzbox' => array('test' => 'value')
+=======
+  'fritzbox' => array( 'uri' => 'https://192.168.0.1:49443/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' ),
+  //'fritzbox' => array( 'uri' => 'http://192.168.0.1:49000/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' ),
+  //'fritzbox' => array( 'uri' => 'https://fritz.box:49443/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' , 'selfsigned' => 'true'),
+  'influx' => array('uri' => 'https://172.17.0.1/proxy/ts/query', 'user' => 'docker', 'pass' => 'cwkvJPQeFlGlRKc402CA', 'selfsigned' => 'true')
+>>>>>>> develop
```
ChristianMayer: This broken as it shows the diff

peuter: Done

ChristianMayer:
- [ ] Not running server (i.e. just running the code as it is) creates a "hidden" 404 error on the `console` instead of giving a nice error/warning to the user about the problem: 
```
Native.js:48 011579 cv.io.rest.Client: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL /git/source/rest/cv/fs was not found on this server.</p>
<hr>
<address>Apache/2.4.25 (Debian) Server at 127.0.0.1 Port 28080</address>
</body></html>
```
- [ ] `sprintf` shows a warning on the console:
```
String.js:72 TypeError: [sprintf] expecting number but found string
    at sprintf.format (sprintf.js:54)
    at wrapper.sprintf (sprintf.js:21)
    at wrapper.cv.util.String.sprintf() (String.js:69)
    at wrapper.cv.ui.structure.pure.Page.prototype.applyFormat() [as applyFormat] (BasicUpdate.js:186)
    at wrapper.cv.ui.structure.pure.Page.prototype.defaultValueHandling() [as defaultValueHandling] (BasicUpdate.js:214)
    at wrapper.cv.ui.structure.pure.Page.prototype.defaultUpdate() [as defaultUpdate] (BasicUpdate.js:298)
    at wrapper.cv.ui.structure.pure.Page.prototype.processIncomingValue() [as processIncomingValue] (Update.js:93)
    at wrapper.cv.ui.structure.pure.Page.prototype.update() [as update] (Update.js:79)
    at wrapper.<anonymous> (Model.js:86)
    at Array.forEach (<anonymous>)
    at wrapper.cv.data.Model.prototype.onUpdate() [as onUpdate] (Model.js:85)
    at wrapper.<anonymous> (Model.js:100)
    at Array.forEach (<anonymous>)
    at wrapper.cv.data.Model.prototype.update() [as update] (Model.js:98)
    at wrapper.cv.io.transport.LongPolling.prototype.handleRead() [as handleRead] (LongPolling.js:149)
    at Direct.js:136 
(2) ["%.1f%%", "0%"]
```

- [ ] grunt fails:
```
$ grunt composer:rest:install
Running "composer:rest:install" (composer) task
/bin/sh: 1: composer: not found
Warning: Task "composer:rest:install" failed. Use --force to continue.

Aborted due to warnings.
```

- [ ] Node fails:
```
$ node source/rest/index.js
internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module '/mnt/timberwolf/var/www/html/git/source/rest/index.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)

```

peuter:
Regarding the last 2 error reports:
grunt fails: As I wrote in a comment here, you have to install composer (the dependency manager for PHP). The `composer` command has to be available in the console.

node fails: the node backend has been completely replaced by the PHP one. There should be no `source/rest/index.js` file and nothing has to be run with node.

ChristianMayer:
Not running anything special (as it's now included in the PHP, I guess) and opening `.../git/source/?manager=1` I get the manager.
On the console error 404 messages are appearing:
Screenshot_20191110_220300
In the manager itself no data is showing up at all.

peuter:
Can you try this URL and check if you get a non 404 response `source/rest/cv/index.php/fs?path=.`
Then the problem is the missing (or deactivated) URL redirection in the webserver.

We can use the URLs without redirection to lessen the server requirements. As a quick fix you can change this line: https://github.com/peuter/CometVisu/blob/feature-restapi/source/class/cv/io/rest/Client.js#L22

`path += 'rest/cv/index.php';` Then it should work without URL redirection.

ChristianMayer:
Changing line 22 to
```
        //path += 'rest/cv';
        path += 'rest/cv/index.php';
```
results in:
Screenshot_20191116_221557

peuter:
Does the `rest/cv/index.php` exist on your server. And if that is the case what would be the correct path to open this file in your server?

ChristianMayer:
- [ ] the manager window should rescale with the browser window
- [ ] the text editor doesn't work (XML editor does show up though)
- [ ] "Prüfen" is just showing the number of errors but not offering a way to see what's wrong
- [ ] "Umbenennen" doesn't work

ChristianMayer:
- [ ] Still necessary to chaninge line 22 to
  ```
        //path += 'rest/cv';
        path += 'rest/cv/index.php';
  ```
- [ ] Download button for a folder at the media section doesn't work - and would only make sense when it would download a full archive (Zip file) of it. I don't think we should add such a complexity to create this archive on the fly => I'd just disable it for a folder
- [ ] Rename media file creates error message

ChristianMayer:
- [ ] Icon: Menu (Hamburger-Icon) doesn't make sense and probably should be deactivated. In the menu: Neue Datei, Neuer Ordner, Wiederherstellen shouldn't be available
- [ ] Saving hidden config at the hidden editor with a 403 (not writeable) response from the server shows 
   ```
   Xhr.js:442 PUT https://timberwolf76.local/proxy/visugit/git/source/rest/cv/index.php/config/hidden 403
   Native.js:48 189838 cv.io.rest.Client:
   ```
  in the console (probably ok) - but also a very wrong feedback to the user: 
grafik
- [ ] After modifying the hidden config with the hidden editor and not saving and then closing the editor doesn't show a warning message that the change wasn't saved
- [ ] Having twice the same key in the hidden editor is during save just overwriting the first key without any warning
- [ ] Text editor: Schreibgeschützter editor doesn't show that it's "special", only when you want to edit text a popup appears (e.g. with an Icon, puting the file name in brackets, ...???). It still contains the menu "Datei->Speichern Strg-S" though
- [ ] The context menu at the tabs can probably be removed or should be cleaned as many entries aren't relevant there: 
grafik
- [ ] Text editor: "Bearbeiten -> Versteckte Konfiguration" opens hidden.php (again and again). That menu entry is surprising and thus shouldn't be there
- [ ] Konfigurationen: Config names should be displayed as they are and not with an uppercase first letter - otherwise "visu_config_t.xml" and "visu_config_T.xml" couldn't be distinguished from each other
- [ ] The icons/buttons for <Default> should be the same as for the other configs: 
grafik
- [ ] Main view: Datei->Datei Schließen doesn't make sense (which file will be closed?)
- [ ] Main view: Datei->Löschen doesn't make sense (which file will be deleted?)
- [ ] Close dialog should show which file (i.e. file name) will be affected: 
grafik
- [ ] Datei->Neu->Neue Datei shows dialog, putting in a name there then this popup shows up 
grafik also no file was created at the file system
- [ ] Konfigurationen: Burger-Menu: "Wiederherstellen" seems to do nothing
- [ ] Konfigurationen: Burger-Menu: "Neuer Ordner" doesn't make sense here

ChristianMayer:
Starting with demo_flat, pressing Strg+M, the console comes up with a few errors on the browser console (first yellow, the others are red):
```
Native.js:48 011871 : Conflicting property method qx.ui.splitpane.Splitter.isVisible with qx.ui.core.Widget
cv.log.appender.Native.process()	@	Native.js:48
qx.log.Logger.__log()	@	Logger.js:603
qx.log.Logger.warn()	@	Logger.js:294
(anonymous)	@	Property.js:612
qx.core.Property.__attachPropertyMethods()	@	Property.js:602
qx.core.Property.attachMethods()	@	Property.js:445
qx.Class.__addProperties()	@	Class.js:1291
qx.Class.define()	@	Class.js:209
(anonymous)	@	Splitter.js:27
```
...
```
Uncaught TypeError: Cannot read property 'getInstance' of undefined
    at wrapper.cv.ui.manager.Main.prototype._onChangeFileSelection() [as _onChangeFileSelection] (Main.js:334)
    at Direct.js:136
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
    at Direct.js:135
    at Array.forEach (<anonymous>)
    at wrapper.qx.event.dispatch.Direct.prototype.dispatchEvent() (Direct.js:120)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js:537)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js:924)
    at Object.qx.event.Registration.__fireNonBubblingEvent() [as __fireNonBubblingEvent] (Registration.js:405)
    at Object.qx.event.Registration.fireNonBubblingEvent() [as fireNonBubblingEvent] (Registration.js:426)
    at wrapper.qx.core.Object.prototype.fireDataEvent() [as fireDataEvent] (MEvent.js:232)
    at wrapper.cv.ui.manager.viewer.Folder.prototype._onSelectionChange() [as _onSelectionChange] (MMultiSelectionHandling.js:436)
    at Direct.js:136
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
    at Direct.js:135
    at Array.forEach (<anonymous>)
```
...
```
Class.js:704 Uncaught Error: Singleton instance of [Class cv.ui.manager.Main] is requested, but not ready yet. This is most likely due to a recursive call in the constructor path.
    at Function.qx.Class.getInstance() [as getInstance] (Class.js:704)
    at wrapper.<anonymous> (Application.js:223)
    at construct.onLoad (Part.js:254)
    at Array.<anonymous> (Part.js:189)
    at construct.qx.Part.prototype.notifyPartResult() [as notifyPartResult] (Part.js:376)
    at construct.qx.io.part.Part.prototype._markAsCompleted() [as _markAsCompleted] (Part.js:202)
    at construct.qx.io.part.Part.prototype.load() [as load] (Part.js:172)
    at Array.onLoad (Part.js:143)
    at construct.qx.Part.prototype.notifyPackageResult() [as notifyPackageResult] (Part.js:431)
    at construct.<anonymous> (Package.js:199)
    at construct.loadScripts (Package.js:232)
    at Package.js:253
```
also a few yellow lines like
```
Bootstrap.js:751 [Violation] 'readystatechange' handler took 234ms
AnimationFrame.js:176 [Violation] 'requestAnimationFrame' handler took 64ms
```

ChristianMayer:
The icon "eye" vs. "lines" is still inconsistent and I can't see any pattern when the one or the other will be shown:
grafik

peuter:
It depends how the file can are opened, when you double-click them: "eye"=>viewer, "lines"= editor.

ChristianMayer:
- [ ] Hidden Config with read-only hidden.php, right click on eye-icon (as well as double click) throws: 
```
Folder.js:182 Uncaught TypeError: ev.getTarget(...).getModel is not a function
    at wrapper.cv.ui.manager.viewer.Folder.prototype._onFsItemRightClick() [as _onFsItemRightClick] (Folder.js:182)
    at EventHandler.js:316
    at Function.qx.event.Utils.series() [as series] (Utils.js:331)
    at EventHandler.js:314
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
    at wrapper.qx.ui.core.EventHandler.prototype._dispatchEvent() [as _dispatchEvent] (EventHandler.js:313)
    at AbstractBubbling.js:276
    at Function.qx.event.Utils.series() [as series] (Utils.js:331)
    at AbstractBubbling.js:262
    at Function.qx.event.Utils.series() [as series] (Utils.js:331)
    at AbstractBubbling.js:257
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
    at wrapper.qx.event.dispatch.AbstractBubbling.prototype.dispatchEvent() [as dispatchEvent] (AbstractBubbling.js:253)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js:924)
    at Registration.js:346
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
```
- [ ] hidden config, read-write, right clicking also throws (but left click works fine now)

ChristianMayer:
- [ ] Im Buger-Menü der einzelnen Config-Dateien/Icons macht es keinen Sinn einen Eintrag für "Neuen Ordner" zu haben. Das passt eher in die Icon-Leiste neben dem "Konfigurationen". Das gleiche gilt für "Neue Datei".
  Hintergrund: beides Bezieht sich eher auf den Ordner, also etwas übergreifendes. Und nicht auf die einzelne Datei.
- [ ] Löschen von Dateien über das Experten-Menü geht nicht, teilweise Fehlermeldung `Document.js:57 Uncaught TypeError: Cannot read property 'nodeType' of null`
- [ ] Wenn ich in der Experten-Spalte auf "media" mit Rechtsklick dort eine neue Datei erzeugen möchte kommt Fehlermeldung `Uncaught TypeError: Cannot read property 'nodeType' of null` (vgl. Gitter-Chat)
- [ ] Rechtsklick auf "media", neuer Ordner -> Ordner wird auf der Hauptebene und nicht unter Media angelegt
    => Zum einen von der Bedienung unlogisch, zum anderen sollte es auf der Hauptebene keine anderen Ordner als "backup" und "media" geben, so dass die Struktur einheitlich und aufgeräumt ist.
- [ ] Demo-Konfigs können nicht mit dem XML-Editor geöffnet werden (weder normal noch Experten-Sicht)
- [ ] Hidden-Config-Editor speichert `' ` ohne Escape und verhindert so die weitere Bearbeitung, egal ob mit Hidden-Config-Editor oder Text-Editor
- [ ] Konfig-Datei ersetzen scheitert mit: 
grafik
    `POST https://timberwolf76.local/proxy/visugit/git/source/rest/cv/index.php/fs?type=file&path=. 406`
    Response Inhalt: `{"message":"Folder already exists"}`

peuter:
I am not able to reproduce the 2 open tasks from you last comment. Please re-test maybe they have been fixed by the last changes too. Otherwise I need more information for reproduction. All other issues should have been addressed now and this PR is ready for another review.

ChristianMayer:
The two points seem to be resolved.

But now I found:
- [ ] Media-Folder (Expert View), add file, then save it without adding anything (i.e. it should be a 0 byte file):
    grafik
- [ ] Further testing: Adding a file(!) in the expert view not only fails with the message above, it even seems to create a folder(!) instead of a file
    Note after further testing: the same issue happens from the normal view; it also happens with the "neue Konfig Datei" which opens a templated new config but then fails as it's creating a folder(!) on the file system with the same name instead
- [ ] Changing write flag for a file on the file system (+w to -w or -w to +w) and then interacting with it on the expert view (like trying to open it for editing, trying to start the rename dialog, ... - i.e. just trying to get to that UI element; usually after hitting the "Neu laden" button) sometimes throws:
```
Main.js:372 Uncaught TypeError: Cannot read property 'isPermanent' of undefined
    at wrapper.cv.ui.manager.Main.prototype.openFile() [as openFile] (Main.js:372)
    at wrapper.cv.ui.manager.Main.prototype.__openSelectedFile() [as __openSelectedFile] (Main.js:277)
    at wrapper.cv.ui.manager.Main.prototype._onChangeTreeSelection() [as _onChangeTreeSelection] (Main.js:258)
    at Direct.js:136
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
    at Direct.js:135
    at Array.forEach (<anonymous>)
    at wrapper.qx.event.dispatch.Direct.prototype.dispatchEvent() (Direct.js:120)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js:537)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js:924)
    at Object.qx.event.Registration.__fireNonBubblingEvent() [as __fireNonBubblingEvent] (Registration.js:405)
    at Object.qx.event.Registration.fireNonBubblingEvent() [as fireNonBubblingEvent] (Registration.js:426)
    at wrapper.qx.core.Object.prototype.fireDataEvent() [as fireDataEvent] (MEvent.js:232)
    at wrapper.<anonymous> (FileSystem.js:175)
    at wrapper.<anonymous> (Timer.js:123)
    at Direct.js:136
```
- [ ] The menu for left and for right mouse button are different: 
grafik vs. 
grafik
    -> I don't understand why the comparison entry is only available with the right button menu
- [ ] in expert, going to the backup folder, right click a file, "Wiederherstellen" doesn't work, the according config file isn't restored (most likely it isn't changed at all)
- [ ] clicking on a folder and then "open" shows an empty screen. When there's nothing that can be done there it'd be better not to offer an open command for a folder:
    grafik
- [ ] Text file in a subfolder under the media folder can not be edited: grafik
- [ ] Creating a new file or config with bad filename (like `test<br/>sdsf`) creates "strange" effects: 
grafik
    -> I think we should be rather strict on the allowed special chars, probably a-z, A-Z, 0-9, " ", "-", "_" and "." is sufficient
- [ ] After deleting a few folders and files in the expert view deleting an other one it throws:
```
Folder.js:254 Uncaught TypeError: Cannot read property 'getFullPath' of null
    at wrapper.cv.ui.manager.viewer.Folder.prototype._handleFileEvent() [as _handleFileEvent] (Folder.js:254)
    at wrapper.qx.event.message.Bus.prototype.__callSubscribers() [as __callSubscribers] (Bus.js:370)
    at wrapper.qx.event.message.Bus.prototype.dispatch() [as dispatch] (Bus.js:302)
    at wrapper.qx.event.message.Bus.prototype.dispatchByName() [as dispatchByName] (Bus.js:326)
    at Function.qx.event.message.Bus.dispatchByName() [as dispatchByName] (Bus.js:114)
    at wrapper.<anonymous> (FileController.js:226)
    at Object.<anonymous> (Client.js:170)
    at Direct.js:136
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
    at Direct.js:135
    at Array.forEach (<anonymous>)
    at wrapper.qx.event.dispatch.Direct.prototype.dispatchEvent() (Direct.js:120)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js:537)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js:924)
    at Registration.js:346
    at Function.qx.event.Utils.then() [as then] (Utils.js:158)
```

peuter:
I am not able to reproduce the last one, maybe its been fixed with the other fixes. All other problems should be fixed.

ChristianMayer:
- [ ] Config: Open with XML Editor, click pages, click page then on the right side click in the "name" field 
    grafik
    Cross check showed that this doesn't happen w/o this PR
- [ ] XML editor: preview doesn't work anymore
- [ ] XML editor: has a button for the manager - which then opens the old manager inside the new manager
- [ ] XML editor: open file and directly save it: 
    grafik
- [ ] XML editor: open file, just change something (like the page name; ignoring the failures of the first point above) then the saved file becomes corrupted:
```
<pages lib_version="8" design="pure" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../visu_config.xsd"><meta><plugins/><mappings/><stylings/><statusbar><status type="html"><img src="resource/icon/comet_64_ff8000.png" alt="CometVisu" /> by <a href="https://www.cometvisu.org/">CometVisu.org</a>
          - <a href="manager.php">Config-Manager</a>
          - <a href=".?forceReload=true">Reload</a></status><status type="html" condition="!edit" hrefextend="config">- <a href="editor/">Edit</a></status><status type="html" condition="edit" hrefextend="config">- <a href=".">normal Mode</a></status><status type="html">- <a href="?config=demo">Widget Demo</a></status><status type="html" hrefextend="config">- <a href="check_config.php">Check Config</a></status></statusbar></meta><page name="StartXXXxx"/></pages>
```
- [ ] Happens mostly sporadic, this time after config file burger menu icon -> compare with older version; closing the comparison window; clicking again on the config burger menu icon:
    grafik
- [ ] even more sporadic after a few clicks (see first line as it didn't appear on the point above): 
    grafik
- [ ] sporadic: open a text file in the text editor in a sub folder of the media folder, open a text file next to it (i.e. two editor tabs are open at the same time), clicking on the first file in the expert view throwed:
```
Document.js?nocache=…9432554448313488:57 Uncaught TypeError: Cannot read property 'nodeType' of null
    at Object.qx.xml.Document.isXmlDocument() [as isXmlDocument] (Document.js?nocache=…9432554448313488:57)
    at wrapper.cv.ui.manager.editor.Worker.prototype.open() [as open] (Worker.js?nocache=0.05380272555702725:47)
    at wrapper.cv.ui.manager.editor.Source.prototype._applyContent() [as _applyContent] (Source.js?nocache=0.6414016770581312:195)
    at wrapper.cv.ui.manager.editor.Source.prototype.resetContent() [as resetContent] (eval at __installFunctionFromCode (Property.js?nocache=…570063348041449:853), <anonymous>:3:957)
    at wrapper.cv.ui.manager.editor.AbstractEditor.prototype._loadFile() [as _loadFile] (AbstractEditor.js?no…8835338349916964:82)
    at wrapper.qx.core.Object.prototype.base() [as base] (Object.js?nocache=0.7222398928473392:178)
    at wrapper.cv.ui.manager.editor.Source.prototype._loadFile() [as _loadFile] (Source.js?nocache=0.6414016770581312:183)
    at wrapper.set (eval at __installFunctionFromCode (Property.js?nocache=…570063348041449:853), <anonymous>:3:1077)
    at wrapper.cv.ui.manager.editor.AbstractEditor.prototype.$$setFileImpl() [as $$setFileImpl] (eval at __installFunctionFromCode (Property.js?nocache=…570063348041449:853), <anonymous>:4:109)
    at wrapper.cv.ui.manager.editor.AbstractEditor.prototype.setFile() [as setFile] (eval at __installFunctionFromCode (Property.js?nocache=…570063348041449:853), <anonymous>:3:20)
    at wrapper.cv.ui.manager.Main.prototype._onChangeFileSelection() [as _onChangeFileSelection] (Main.js?nocache=0.8273194476456518:329)
    at Direct.js?nocache=0.…673490641217157:136
    at Function.qx.event.Utils.then() [as then] (Utils.js?nocache=0.7300205815097147:158)
    at Direct.js?nocache=0.…673490641217157:135
    at Array.forEach (<anonymous>)
    at wrapper.qx.event.dispatch.Direct.prototype.dispatchEvent() (Direct.js?nocache=0.…673490641217157:120)
```
- [ ] then closing the files (tabs) (i.e. w/o reload that I usually do as soon as I get a throw to be sure to be clean)
```
Start.js?nocache=0.8174433772120933:73 Uncaught TypeError: Cannot read property 'load' of null
    at wrapper.cv.ui.manager.Start.prototype._loadRoot() [as _loadRoot] (Start.js?nocache=0.8174433772120933:73)
    at wrapper.cv.ui.manager.Start.prototype.resetFile() [as resetFile] (eval at __installFunctionFromCode (Property.js?nocache=…570063348041449:853), <anonymous>:3:894)
    at Object.qx.core.Property.__unwrapFunctionFromCode() [as __unwrapFunctionFromCode] (Property.js?nocache=…570063348041449:820)
    at Object.qx.core.Property.executeOptimizedSetter() [as executeOptimizedSetter] (Property.js?nocache=…570063348041449:995)
    at wrapper.members.<computed> (Property.js?nocache=…570063348041449:682)
    at wrapper.cv.ui.manager.Main.prototype.closeFile() [as closeFile] (Main.js?nocache=0.8273194476456518:429)
    at wrapper.<anonymous> (Main.js?nocache=0.8273194476456518:406)
    at wrapper.dialog.Confirm.prototype._handleYes() [as _handleYes] (Confirm.js?nocache=0…436263522242273:148)
    at Direct.js?nocache=0.…673490641217157:136
    at Function.qx.event.Utils.then() [as then] (Utils.js?nocache=0.7300205815097147:158)
    at Direct.js?nocache=0.…673490641217157:135
    at Array.forEach (<anonymous>)
    at wrapper.qx.event.dispatch.Direct.prototype.dispatchEvent() (Direct.js?nocache=0.…673490641217157:120)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js?nocache…141421720257341:537)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js?nocache=0…946637852406233:924)
    at Registration.js?noca…138868596155904:346
```
- [ ] then replacing the file with a new one in the expert view (note: after a restart a replacement of the file worked fine):
```
Folder.js?nocache=0.4394153749631944:261 Uncaught TypeError: Cannot read property 'getFullPath' of null
    at wrapper.cv.ui.manager.viewer.Folder.prototype._handleFileEvent() [as _handleFileEvent] (Folder.js?nocache=0.4394153749631944:261)
    at wrapper.qx.event.message.Bus.prototype.__callSubscribers() [as __callSubscribers] (Bus.js?nocache=0.009379649789500677:370)
    at wrapper.qx.event.message.Bus.prototype.dispatch() [as dispatch] (Bus.js?nocache=0.009379649789500677:302)
    at wrapper.qx.event.message.Bus.prototype.dispatchByName() [as dispatchByName] (Bus.js?nocache=0.009379649789500677:326)
    at Function.qx.event.message.Bus.dispatchByName() [as dispatchByName] (Bus.js?nocache=0.009379649789500677:114)
    at wrapper.<anonymous> (UploadMgr.js?nocache…676588345192057:121)
    at Direct.js?nocache=0.…673490641217157:136
    at Function.qx.event.Utils.then() [as then] (Utils.js?nocache=0.7300205815097147:158)
    at Direct.js?nocache=0.…673490641217157:135
    at Array.forEach (<anonymous>)
    at wrapper.qx.event.dispatch.Direct.prototype.dispatchEvent() (Direct.js?nocache=0.…673490641217157:120)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js?nocache…141421720257341:537)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js?nocache=0…946637852406233:924)
    at Registration.js?noca…138868596155904:346
    at Function.qx.event.Utils.then() [as then] (Utils.js?nocache=0.7300205815097147:158)
    at Object.qx.event.Registration.fireEvent() [as fireEvent] (Registration.js?noca…138868596155904:345)
```

peuter:
I assume that the last two are just subsequent errors. I fixed the `cache=…9432554448313488:57 Uncaught TypeError: Cannot read property 'nodeType' of null` error and assume that the other ones do not occur anymore.

ChristianMayer:
Opening a normal config (demo_flat), typing Ctrl+M the manager with an empty main screen opens and also throws an assertion on the console:
grafik
Note: the same problem happens when the hidden.php is writeable, i.e. it's not showing the error popup

ChristianMayer:
Normal config, XML-Editor: preview fails with 404:
grafik

In both cases it the error popup content not useful for the user.
Normal save fails with 405:
grafik

peuter:
Regarding the ` Media, (sub)folder: file name is croped away:` => All file names in the simple view do not use the real name: Configs are only shown without the `visu_config<>.xml` part, the special files like hidden.php os the icons use another name, or even do not exists in reality. And the file endings are cropped away on purpose, because the space is limited and the file type should be recognizable by the icon used there. If the user wants to use exotic and long file endings he must use the expert mode to see the complete name. I added the full filename to the tooltip of the file icons, but will not change the cropping.

peuter:
Regading the broken hidden config: I did not intent so solve every possible problem with this first shot.
It is planned to add section templates to the hidden config, that define a structure (what options are available and so on) that should also include some kind of syntax check. But thats for the future. Currently the user still can open the hidden config in the text editor to fix the broken structure.
I guess it would be hard to add a syntax check for php in pure javascript and I don't want to add it to the backend, because the rest backend should stay as stupid as possible. And although it would be easy to check the syntax in the php backend, it would be harder in the java backend in openHAB.
For better syntax checking possibilties we should consider and programming language independent syntax, like JSON for example, together with JSON-Schema the syntax checking would be quite easy. But again => not the purpose of this PR but for future improvements.

peuter:
I will look for errors about the XML-Editor saving next time. The old editor is a touch one to integrate in the new manager. Should re re-implemented in the long run => future task. But it should of course work without errors, will have a look at it next weekend (hopefully)

peuter:
The config preview error should be gone (file path for preview was wrong). 
I am not able to reproduce the the 405 error when saving the file, saved several times => no error. If you can still reproduce this error please provide more information on how to run into that error.

All other topics should have been addresses now. If I forgot something, please let me know.

ChristianMayer:
- [ ] The file system structure should be enforced, but as that could be done in a 2nd step I put that point also at https://github.com/CometVisu/CometVisu/issues/973#issuecomment-599137312
- [ ] 405 is still happending (see Gitter chat https://gitter.im/CometVisu/CometVisu_DE?at=5e6d4aac1cf20f12f95f0583)
- [ ] after some fiddling this error is shown (doesn't seem to affect the manager negatively): 
grafik
- [ ] renaming a file with a name that contains a `/ ` fails in a "strange" way 
grafik
- [ ] sometimes (probably when writing a file that is not writeable?) files with datecode in the name get created (in my example case `media/visu_config_tetetetetetetet.xml` -> `media/visu_config_tetetetetetetet.xml260461733`).

PR #948, opened 2019-06-26
fix the problem that slider updates are send back to the backend agai…

peuter:
…n, because the changeValue event handler is throttled and got executed delayed (when `this.__skipUpdatesFromSlider` was already set back to false).

This commit changes that behaviour by checking `this.__skipUpdatesFromSlider` outside the throttled handler.

### DIFF
--- a/source/class/cv/ui/structure/pure/Slide.js
+++ b/source/class/cv/ui/structure/pure/Slide.js
@@ -112,8 +112,13 @@ qx.Class.define('cv.ui.structure.pure.Slide', {
         slider.init();
         // set initial value
         slider.setValue(parseFloat(this.getValue()));
+        var throttled = qx.util.Function.throttle(this._onChangeValue, 250, {trailing: true}).bind(this);
 
-        slider.on("changeValue", qx.util.Function.throttle(this._onChangeValue, 250, {trailing: true}), this);
+        slider.on("changeValue", function (ev) {
+          if (!this.__skipUpdatesFromSlider) {
+            throttled(ev);
+          }
+        }, this);
 
         this.addListener("changeValue", function (ev) {
           this.__skipUpdatesFromSlider = true;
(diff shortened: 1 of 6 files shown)


peuter:
> The changes in `.travis.yml` shouldn't be part of this PR

Without these changes the pr has no chance to be build successfully (because of the chromedriver not supporting current chrome unstable). The remaining failing test seems to be a real error which I haven't been able to fix yet (or even understand whats going on). Due to the lack of time, it might take some time to figure it out and solve it.

ChristianMayer:
OK about the `.travis.yml
But a real failing test prevents a merge, as it would mean that all future PRs would also fail the tests.

ChristianMayer:
At the moment my IDE doesn't work at reproducing this error - it fails much more early, I guess on some setup issues.
=> I can't help here at the moment.

Even when the source of the blocking issue is found I think the involved test is quite fragile, as the output currently says:
```
Expected 459.171875 to equal 409.671875.
```
General remark:
A test for equality of floats (`expect(newPos.x).toEqual(...)` will easily fail due to numerics. Please consider to change that to a `toBeCloseTo()` test.

peuter:
Tests are ok now.

peuter:
Ok I tried to activate the chrome unstable tests, but that did not work. There might be a (complicated) way to make it work, but that is not reliable because the chrome driver is only available for beta and stable. For unstable there might by one available but not always (Quote from: http://chromedriver.chromium.org/downloads/version-selection
```
We always provide ChromeDriver for the current Stable and Beta versions of Chrome. However, if you use Chrome from Dev or Canary channel, or build your own custom version of Chrome, It is possible that there is no available ChromeDriver that officially supports it.
```
So I suggest to remove the unstable tests completely, beta and stable should be enough to find the problems early enough.

peuter:
The unit tests are still executed for chrome unstable only the e2e tests are disabled.

As soon as the build is green this PR is ready to be merged.

source/class/cv/ui/structure/pure/Slide.js
```diff
@@ -112,7 +112,7 @@ qx.Class.define('cv.ui.structure.pure.Slide', {
         slider.init();
         // set initial value
         slider.setValue(parseFloat(this.getValue()));
-        var throttled = qx.util.Function.throttle(this._onChangeValue, 250, {trailing: true}).bind(this);
+        var throttled = qx.util.Function.throttle(this._onChangeValue, 250, {trailing: false}).bind(this);
```
ChristianMayer: `{trailing: true}` (or something similar?) is required to not break the user action:

When dragging the slider quickly out of it's range (e.g. to force it to 0% or 100%) will not work anymore with this change

peuter:
OK next try, not with `{trailing: true}`. Error is fixed by using an own implementation of the `throttle`function which is aborted when the user stops dragging the slider.

Issue #772, opened 2018-12-13
Minimize 3rd party library lock in


ChristianMayer:
The source code should use as much native libraries (-> W3C) and as little 3rd part libraries (Qooxdoo, JQuery) as possible and sensible.

Reasoning: This will make future library changes more easy.

To get closer to this target requires personal judgement at each library use. Good guides are:
- is that feature (e.g. modifying attributes of a DOM element) also possible with a browser native library?
- is the native library well implemented in many browsers (-> caniuse)?
- is it also implemented in older browsers (plugins are allowed to use bleeding edge, core code must stay conservative)? Desktop browsers can be assumed to get updated but mobile browsers might not

peuter:
Well I don't see that much potential here, lets see what we use:
* Qooxdoo: cannot be replaced unless someone wants to rewrite evenrything again. So it makes no sense to avoid using Qooxdoo inside the source code, as its the basis of everything
* EventRecorder.js: needed for log recoding and is no 3rd party library, because I wrote it
* sprintf.js: Used for mappings AFAIR. Might be possible to replace it, but if you want to keep the same syntax you have to use another library or write it yourself => both options provide no benefit
* JQuery: not used by the core anymore. Only some plugins still need it. We should take care that new plugins do not rely on jquery and in the very long run remove the jquery dependency from the remaining plugins. But this has a very low priority for me.
* strftime.js: Could be replaced with Qooxdoos own Date/Timeformatting skills. but that would imply a syntax change
* svg4everybody.js: Not really sure but AFAIR this is needed for SVG-Support in some IE Browsers. Might be obsolete in future (or even now)
* favico.js: Used to change the Favicon dynamically (e.g. show how many (error) messages are in the Notification center. Can't be replaced without using that feature.

ChristianMayer:
This issue is clearly long term. And it's not about writing everything our self. It's about keeping the ability to choose in the future. So it's purely strategic.

------

- To throw out libraries I see most potential at the remaining JQuery use.
- To reduce lock in I see most potential at Qooxdoo. E.g. DOM handling in the browsers has improved significantly since the CV has started. Anything that's easily(!) possible native should be done native and not by the Qooxdoo wrappers. Anything that's not easily possible, well, that's the reason we are using a library, isn't it?

peuter:
Well, using libraries only when they provide some kind of benefit (performance, easier to use, additional features, etc.) should be obvious.

The plugins that still use jquery are:

- [ ] Openweathermap
[...]

PR #953, opened 2019-09-15
Reduce some Qx and switch to native

ChristianMayer:
See also #772

This patch reduces the vendor lock in to Qx in about 1000 instances where native JavaScript does the job exactly as well as the Qx layer on top of it.

So this patch means that for each of those actions less code is being run (should give a minimal speed advantage), the code is (at least uncompressed) smaller by about 3.6kb and - most importantly - helps us to stay as independent from any unnecessary code as possible.

There are still many places that would benefit from such a clean up, but it must start somewhere.

### DIFF
--- a/skeletons/widget-test.js
+++ b/skeletons/widget-test.js
@@ -6,7 +6,7 @@ describe("testing a %WIDGET_NAME% widget", function() {
 
   it("should test the %WIDGET_NAME% creator", function() {
     var res = this.createTestWidgetString("%WIDGET_NAME%", {}, "<label>Test</label>");
-    var widget = qx.bom.Html.clean([res[1]])[0];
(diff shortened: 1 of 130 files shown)


skeletons/widget-test.js
```diff
@@ -6,7 +6,7 @@ describe("testing a %WIDGET_NAME% widget", function() {
 
   it("should test the %WIDGET_NAME% creator", function() {
     var res = this.createTestWidgetString("%WIDGET_NAME%", {}, "<label>Test</label>");
-    var widget = qx.bom.Html.clean([res[1]])[0];
+    var widget = (function(){var div=document.createElement('div');div.innerHTML=res[1];return div.childNodes[0];})();
```
peuter: As it is not really clear what this function does, it should be declared somewhere outside the tests with a self explaining name and a little description. That function can be used everywhere, where `qx.bom.Html.clean`got replaced.

PR #984, opened 2020-01-18
Client robustnes against temporary glitches

ChristianMayer:


client/source/class/cv/io/Client.js
```diff
@@ -491,7 +492,8 @@ qx.Class.define('cv.io.Client', {
           }
           if (options.listeners) {
             Object.getOwnPropertyNames(options.listeners).forEach(function(eventName) {
-              ajaxRequest.addListener(eventName, options.listeners[eventName], context);
+              var qxEventName = 'error' !== eventName ? eventName : 'statusError';
+              ajaxRequest.addListener(qxEventName, options.listeners[eventName], context);
```
peuter: The 'statusError' listener is always added to ajaxRequest (see some lines below this one). What is the reason for this change? IMHO this change would result in two listeners to the 'statusError' event and no listener to the 'error' event.

client/source/class/cv/io/transport/LongPolling.js
```diff
@@ -208,6 +211,14 @@ qx.Class.define('cv.io.transport.LongPolling', {
...
       "qx": function (ev) {
         var req = ev.getTarget();
+        // check for temporary server errors and retry a few times
+        if([408,444,499,502,503,504].indexOf(req.getStatus()) >= 0 && this.retryServerErrorCounter < this.client.backend.maxRetries) {
+          this.info('Temporary connection problem (status: ' + req.getStatus() + ') - retry count: ' + this.retryServerErrorCounter);
+          this.retryServerErrorCounter++;
+          req.serverErrorHandled = true;
+          this.restart();
+          return;
+        }
```
peuter: The retry-handling feature should also be added to the jquery version of the client. Otherwise the two client releases would behave differently.

config.json
```diff
@@ -296,6 +296,7 @@
       {
         "cv.build": "source",
         "cv.sentry": false,
+        "qx.dynlocale": false,
```
peuter: Deactivating the dynamic locale switching is fine, as we do not need it in the client. But out of curiosity: Did you experience any problems with this feature here, that made you change the setting?

client/source/class/cv/io/Client.js
ChristianMayer: That's the other way round:
This line says: "please use `eventName` - except when it's `'error'`, then replace that with `'statusError'`"

There can be many listeners to `statusError`, one from the Client itself (line 507) and one per `options.listeners` at this line.

The reason for that change was that the listener for the `error` wasn't called, even when I made the server to give back a few 502 status codes

config.json
ChristianMayer: Yes, there were some problems (cf. Gitter conversation).
The code is expecting a String in some places but only got the translation-object instead.

client/source/class/cv/io/transport/LongPolling.js
ChristianMayer: As I have no way to test as the moment I didn't want to take the risk for such a change at a very critical key component of the CV.
The PR is "just" improving robustness, so I think it's ok to have a slightly different behaviour

peuter: Ok, I assume that there are not many (or even no) users for the standalone client in the jquery version, I guess we can live with that difference.

PR #990, opened 2020-01-26
Feature collapsable navbars

ChristianMayer:
Make the navbars dynamically collapsable.

Note: this should - at least partially - replace the mobilemenu plugin.

Just review and don't pull yet as I had to remove a test and this should be replaced by a new test.
(Due to the enhanced functionality it should be a few more tests)

### DIFF
--- a/source/class/cv/ui/structure/pure/NavBar.js
+++ b/source/class/cv/ui/structure/pure/NavBar.js
@@ -56,8 +56,9 @@ qx.Class.define('cv.ui.structure.pure.NavBar', {
       init: 'left'
     },
     dynamic: {
-      check: "Boolean", 
-      init: false
+      check: [true, false, null],
+      nullable : true,
+      init: null
     },
     visible: {
       refine: true,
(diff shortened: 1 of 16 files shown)


source/resource/designs/metal/basic.css
```diff
@@ -587,7 +587,11 @@ div#demo_3 {
 	width: auto;
 }
 
-#navbarTop {
+#navbarTop,
+#navbarBottom,
+#navbarLeft,
+#navbarRight
+{
```
peuter: This changes the left/right navbar background to the gradient. Left/Right navbars in metal design do not have a gradient background only the top/bottom ones.

source/class/cv/ui/structure/pure/NavBar.js
```diff
@@ -56,8 +56,9 @@ qx.Class.define('cv.ui.structure.pure.NavBar', {
       init: 'left'
     },
     dynamic: {
-      check: "Boolean", 
-      init: false
+      check: [true, false, null],
```
peuter: This change is not needed, the "Boolean" check allows `null` values when `nullable` is `true`.

ChristianMayer:
PR should be ready now

PR #992, opened 2020-03-07
only use the content of real text nodes, e.g. avoid using the textual…

peuter:
… content of comments

found during investigation of https://knx-user-forum.de/forum/supportforen/cometvisu/1474911-neuinstallation-von-cv-0-11-2-auf-raspi-was-muss-alles-installiert-werden/page3#post1478000


peuter:
The fix for the failing travis job is part of #947  and should be ignored here

PR #997, opened 2020-03-17
upgrade multitrigger to support multiple buttons and make the linefee…

peuter:
…d configurable


peuter:
demo configs are updated

PR #999, opened 2020-03-29
Feature: roundbar widget

ChristianMayer:
This PR adds a new widget - the roundbar.
Also a few minor fixes on other places.


peuter:
This really looks great, I think we should add some fancy looking examples to te CometVisu-Demo page https://www.cometvisu.org/CometVisu/de/latest/demo/.

Issue #1004, opened 2020-04-12
Enhance documentation: reference setups for developers


ChristianMayer:
The documentation should contain as a help reference setups for developers.
The use cases should be at least:

- tester: Windows or Mac computer, no dedicated develpment tools and no intention to install them, visu runs on a external linux server or a docker image with little to none access; wants only to change files when told so by a developer ("please try to change line X in file Y to Z") and needs to report back (with instructions) browser internal data
- casual developer: Windows computer (no development tools installed yet and only wants to install something light weight), visu runs on a Linux computer or Linux docker image with some access; only wants to write a plugin or do some little code tweaks
- advanced developer: deeper development know how although probably not with ECMAScript. Runs Windows or Linux. Has already an IDE or a powerful text editor. Is prepared to do some deeper installations. Visu runs on a Linux computer or Linux docker image with the possibility to get the necessary access.
- hard core developer: System will be set up as necessary. Visu runs on a Linux computer or Linux docker image with all required access. Focus is ease of development, nothing should be quirky or feel not smooth.

PR #1007, opened 2020-04-13
development setup documentation - NOT FINISHED YET

ChristianMayer:
Solves #1004


peuter:
This mightl also be interesting in this context:
https://code.visualstudio.com/docs/remote/remote-overview

PR #1009, opened 2020-04-17
Fix rest api

peuter:
The is the fix collection that arised during development of the openHAB backend for the manager.
Mainly its a sync between the openapi-spec and the reality and some fixes inside the manager.

This also includes a change for the internal data handling of the hidden config (to JSON). Should be transparent to the parts using the hidden config. But is a lot easier to read/write in PHP and JAVA. Also this sets the ground for JSON-Schema validation for the hidden config.

### DIFF
--- a/source/resource/config/hidden.php
+++ b/source/resource/config/hidden.php
@@ -1,9 +1,16 @@
 <?php
-// File for configuraions that shouldn't be shared with the user
-$hidden = array(
-  'fritzbox' => array( 'uri' => 'https://192.168.0.1:49443/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' ),
-  //'fritzbox' => array( 'uri' => 'http://192.168.0.1:49000/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' ),
-  //'fritzbox' => array( 'uri' => 'https://fritz.box:49443/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' , 'selfsigned' => 'true'),
-  'influx' => array('uri' => 'https://172.17.0.1/proxy/ts/query', 'user' => 'docker', 'pass' => 'cwkvJPQeFlGlRKc402CA', 'selfsigned' => 'true')
-);
-?>
+// File for configurations that shouldn't be shared with the user
+$data = '{
+  "fritzbox": {
+    "uri": "https://192.168.0.1:49443/", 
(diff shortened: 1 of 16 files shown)


source/resource/config/hidden.php
```diff
@@ -1,9 +1,16 @@
...
-// File for configuraions that shouldn't be shared with the user
-$hidden = array(
-  'fritzbox' => array( 'uri' => 'https://192.168.0.1:49443/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' ),
-  //'fritzbox' => array( 'uri' => 'http://192.168.0.1:49000/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' ),
-  //'fritzbox' => array( 'uri' => 'https://fritz.box:49443/', 'user' => 'CometVisuTestUser', 'pass' => 'cvtu4here' , 'selfsigned' => 'true'),
-  'influx' => array('uri' => 'https://172.17.0.1/proxy/ts/query', 'user' => 'docker', 'pass' => 'cwkvJPQeFlGlRKc402CA', 'selfsigned' => 'true')
-);
-?>
\ No newline at end of file
+// File for configurations that shouldn't be shared with the user
```
ChristianMayer, line 2: Add marker to show `hidden.php` type version.

E.g.:
```
<?php
// hidden.php type version: 2
// File for configurations that shouldn't be shared with the user
...
```

peuter: Both backends can read the old version, but save the new one. So any modified data will be automatically converted into the new structure on the first change over the REST API.
No other service should directly modify the internal data structure of the hidden.php, thats like writing to an private variable => nogo.
Write operations should always use the REST API. Is that possible with the Timberwolf?

source/rest/manager/src/Api/ConfigApi.php
```diff
@@ -32,39 +32,12 @@ protected function load()
...
-        }
-      $out .= ')';
-    }
-    $out .= "\n);\n?>";
-
-    print_r($this->hidden);
-
-
+// File for configurations that shouldn\'t be shared with the user
+$data = \'' . json_encode($this->hidden, JSON_PRETTY_PRINT) . '\';
```
ChristianMayer, line 38: This should also have the option `JSON_UNESCAPED_SLASHES` to keep configured URLs readable (and probably also `JSON_PRESERVE_ZERO_FRACTION` to prevent modification of data that is not intended by the user, just in case)

PR #1025, opened 2020-05-17
replace links in configs

peuter:
* replace config links for opening manager, editing config file and config file validation
* fix error in text editor not showing errors in readOnly files (e.g. demo configs)
* change toast messages text color with normal severity to black for better readability
* compability layer for existing configs that replaces the old manager links with the new ones

### DIFF
--- a/source/resource/config/.templates/visu_config.xml
+++ b/source/resource/config/.templates/visu_config.xml
@@ -10,11 +10,11 @@
     <statusbar>
       <status type="html"><![CDATA[
           <img src="resource/icon/comet_64_ff8000.png" alt="CometVisu" /> by <a href="https://www.cometvisu.org/">CometVisu.org</a>
-          - <a href="manager.php">Config-Manager</a>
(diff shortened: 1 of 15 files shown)


source/resource/config/.templates/visu_config.xml
```diff
@@ -14,7 +14,7 @@
           - <a href=".?forceReload=true">Reload</a>
         ]]></status>
       <status type="html" condition="!edit" hrefextend="config"><![CDATA[
-          - <a href="editor/">Edit</a>
+          - <a href="#" onclick="qx.core.Init.getApplication().showManager('open', 'visu_config.xml')">Edit</a>
```
ChristianMayer, line 17: The file name shouldn't be hard coded here, especially not for the template. This will lead to questions / bug reports when the user is creating a config with a different name

peuter: Fixed

peuter:
Ping

PR #1028, opened 2020-05-31
Fix broken diagram widget

ChristianMayer:
Fix broken diagram widget due to missing asset.
Improve performance (prevent time out) when a big InfluxDB is used with many tags

### DIFF
--- a/source/editor/dataproviders/list_all_influxdbs.php
+++ b/source/editor/dataproviders/list_all_influxdbs.php
@@ -98,40 +98,15 @@ foreach( $databases[ 'results' ][ 0 ][ 'series' ][ 0 ][ 'values' ] as $databaseE
   $resSeries = array();
   $measurements = array();
 
-  $seriesArr = json_decode( query( 'SHOW SERIES', $database, $_GET['auth'] ), true );
+  $seriesArr = json_decode( query( 'SHOW MEASUREMENTS', $database, $_GET['auth'] ), true );
   $series = $seriesArr[ 'results' ][ 0 ][ 'series' ][ 0 ][ 'values' ];
   if( NULL != $series )
   {
     foreach( $series as $thisSeries )
     {
-      $list = explode( ',', $thisSeries[ 0 ] );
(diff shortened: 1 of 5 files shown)


source/editor/dataproviders/list_all_influxdbs.php
```diff
@@ -98,40 +98,15 @@ function query( $q, $db = '', $auth )
...
-
-      // now forget all the nice information and compact to the relevant one:
       $arrData[] = array(
-        'value' => $database . '/' . $measurement,
-        'label' => $database . '/' . $measurement
+        'value' => $database . '/' . $thisSeries[ 0 ],
+        'label' => $database . '/' . $thisSeries[ 0 ]
       );
     }
   }
```
peuter, line 112: This Dataprovider is deprecated und will not be used by the new manager. Please apply those changes to `source/rest/manager/src/Api/DataproviderApi.php`.

ChristianMayer: I applied the same change to `source/rest/manager/src/Api/DataproviderApi.php` now - but I couldn't test it there as my code uses `source/editor/dataproviders/list_all_influxdbs.php`.

peuter: I can't test it either because I do not have an setup with influxdb. but you can just replace the URL with `source/rest/manager/index.php/data/influxdbs` and must see the same result. It should be tested, before we merge it because it seems to be broken anyways. I just wanted to open the URL with the current develop-version and  get an error: `Call to undefined function OpenAPIServer\Api\query()` maybe you can have a look at it.

ChristianMayer: The PHP namespace seems to do strange things (probably it's working fine but as I'm not used to them it looks rather strange to me)

A fix is commited. I don't like the necessary switch from `require_once(hidden.php)` to `require`. But I had no way to make `$hidden` visible inside of the `query` function...

PR #1029, opened 2020-06-07
Manager improvements

peuter:
* remove old links to config checking and config upgrading
* implement a client side config upgrader (only from version =>7) that shows the changes in the diff editor
* open text editor with validation errors if the validation failed
* fix automatic library version  check when a config is loaded
* another compiler upgrade + fixes in tests to avoid accessing private members
* upgrade all lib_versions references the the current one (9)
* strip down the build by not copying the monaco editor dev files
* add some missing `;` in the compiler files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -380,9 +380,9 @@
       }
     },
     "@qooxdoo/compiler": {
-      "version": "1.0.0-beta.20200515-1215",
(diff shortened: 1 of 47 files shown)


ChristianMayer:
Opening an `lib_version="8"` config shows the dialog:
grafik
then selecting "starten ohne Upgrade" fails as the new URL `/?config=test_slider&design=metal&forceReload=true#id_&libraryCheck=false` is broken (the part `&libraryCheck=false` must be before the `#`)

Issue #902, opened 2019-03-16
knxd backend handler might fail


ChristianMayer:
- [ ] A badly configured docker container (e.g. bad KNX tunnel port) creates an error message that the user might not understand and thus can not help himself
- [ ] for whatever reason a request with "i=unknown" could happen (see https://knx-user-forum.de/forum/supportforen/cometvisu/1340421-cometvisu-11-0-portainer-timberwolf-protokollfehler-backend-invalid-response?p=1341162#post1341162 )

asander81:
In my case the reason for this error is, that the answer from backend is "{'error': 'Read failed'}". So there is something going wrong in eibread-cgi.

ChristianMayer:
@asander81 I guess that you are using a knxd newer than version 0.0.5.1? Just knowing about the latest knxd version (I never checked any in between) I found a few issues to be aware of: https://github.com/knxd/knxd/issues/494, https://github.com/knxd/knxd/issues/495 and https://github.com/knxd/knxd/issues/493
A `{'error': 'Read failed'}` is coming from the knxd and eibread-cgi/eibwrite-cgi, so it's not a topic for this issue here.

ChristianMayer:
Current level is fine for 0.12.0, any additional optimization should come for a later release

ChristianMayer:
As we are using a completely new knxd backend, closing this issue now

PR #1036, opened 2020-12-13
Fix error pop up and rsslog

ChristianMayer:
This pull request fixes:
- error pop up is showing a huge number of "neu laden"
- a bad backend response during start up (e.g. a 404 page with HTML text instead of a JSON for "l") was giving a bad hint (only a "{}" response text instead of the real one)
- the error pop up might sometime end up half way out of the left side of the screen
- the rsslog didn't work anymore as the PHP file was missing in the compilation
- the rsslog popup looked broken as the icons were too big


ChristianMayer:
This solves the first point of #902

peuter (approved):

PR #1041, opened 2021-01-18
Enable diagram influx to switch of time grouping

### LINKED ISSUE #1040: Diagram with Influx: call without resolution is not possible
It is not possible to create an influx backed diagram with stored datapoints only, currently always a grouping by time (default: 300s) is happening.

Cf. e.g. https://forum.timberwolf.io/viewtopic.php?p=29626#p29617

ChristianMayer:
Fixes #1040

### DIFF
--- a/source/class/cv/plugins/diagram/AbstractDiagram.js
+++ b/source/class/cv/plugins/diagram/AbstractDiagram.js
@@ -254,7 +254,7 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
      * @param force {Boolean} Update even when the cache is still valid
      * @param callback {Function} call when the data has arrived
      */
-    lookupTsCache: function(ts, start, end, res, refresh, force, callback, callbackParameter ) {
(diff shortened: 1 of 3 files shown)


source/class/cv/plugins/diagram/AbstractDiagram.js
```diff
@@ -367,6 +376,10 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
       check: "Number",
       init: 300
     },
+    forceNowDatapoint: {
```
peuter, line 379: Is this property configurable somehow? I didn't find anythingso it looks like this property is currently obsolete, or is this a preparation for a later change?

ChristianMayer: At the moment it's only using the default value as that's what most use cases will want to use.
But it could be a config option in future where it is changeable.

As it's default value is fixing (i.e. changing) the current behavior of the diagram display I opted to make it easily modifiable for possible future use.

peuter (approved):

PR #1042, opened 2021-01-30
Native openHAB REST API client

peuter:


source/class/cv/plugins/diagram/AbstractDiagram.js
```diff
@@ -285,7 +298,7 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
             this.cache[ key ].xhr.dispose();
           }
           var xhr = new qx.io.request.Xhr(url);
-          var self = this;
```
ChristianMayer, line 288: During compile I get
```
cv.plugins.diagram.AbstractDiagram: [305,50] warning: Unresolved use of symbol self._onSuccessiagram
cv.plugins.diagram.AbstractDiagram: [306,54] warning: Unresolved use of symbol self._onStatusError
```
Reinserting `var self = this;` fixed that

ChristianMayer (changes requested):

source/class/cv/plugins/diagram/AbstractDiagram.js
peuter: We can use this here, because the context is provided to the listeners.

ChristianMayer:
The `package-lock.json` has some conflict between our systems, I guess. I remember that mine once changed all links from `http` to `https` and this PR tries to change it back to `http` again.

For these static files I don't care much about TLS or not (although with TLS is preferred, of course). But we should make it consistent so that it's not changing all the time.

peuter:
Even a complete deletion of node_modules and an new `npm install` did not change the http links in the package-lock.json. So I changed them to https by using sed (according to https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818/3).

source/class/cv/TemplateEngine.js
```diff
@@ -507,7 +531,7 @@ qx.Class.define('cv.TemplateEngine', {
       this.debug("setup");
 
       // login to backend as it might change some settings needed for further processing
-      this.visu.login(true, function () {
+      this.visu.login(true, cv.Config.configSettings.credentials, function () {
```
ChristianMayer, line 534: This breaks `client/source/class/cv/io/Client.js` line 367:
```
login : function (loginOnly, callback, context) {
```
where only 3 and not 4 parameters are expected

ChristianMayer (changes requested):

ChristianMayer (approved):

PR #1047, opened 2021-02-13
add address child to image to allow image URLs to be send via group a…

peuter:
…ddress

### DIFF
--- a/doc/manual/de/config/widgets/image/index.rst
+++ b/doc/manual/de/config/widgets/image/index.rst
@@ -16,6 +16,14 @@ Beschreibung
 
 Das Image Widget fügt der Visualisierung ein statisches Bild oder ein Kamerabild hinzu.
 
+Es gibt zwei Möglichkeiten die URL des Bildes anzugeben. Einmal über den ``src``-Parameter,
+oder als Alternative über den Wert einer Gruppenadresse.
+
+.. hint::
+
+    Die Möglichkeit Bilder über den Wert einer Gruppenadresse zu übermitteln wurde in
+    Version 0.12 eingeführt.
+
 Das Bild muss als URL angegeben werden. Wenn man Dateien anzeigen möchte müssen diese über
 http erreichbar auf dem CV-System oder einem anderen Server liegen.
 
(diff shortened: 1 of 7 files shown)


ChristianMayer:
First few comments:
- (just by looking at the conde, not trying it) isn't `display:none` a bit too harsh as it destroys the layout and causes a reflow? Probably makinng it opaque or replacing the href with an unset address with a 1x1 px transparent GIF or PNG might be less invasive
- KNX can have long texts (IIRC DPT24.001 and DPT28.001), but noone has implemented those yet. Not even the ETS...  
   But even then using a map should work, shouldn't it?

peuter:
Maybe we should add an option to let the user decide what should happen if there is no picture to show? For my usecase it is useful to show nothing when there is nothing to show. The alternative with the current code is a broken image with no dimensions that also breaks the layout and also looks broken.
It s about showing a covert-art when playing a song. Openhab delivers those images as data-url. So the image is inside the string and directly shown with this changes.

What Do you mean with "a map should work"? Having a pre-defined map of image urls where you choose from by sending the key/index of a particular image? That would a) nor work with the use-case described above (data-url) and b) won't work when you have an unknown set if images (e.g. showing a cover-art from a currently played song via a streaming services with a huge amount of possible cover arts.

ChristianMayer:
I fully understand the use case where you want to display no picture at all. But I recon a `visibility: hidden` (or a `opacity: 0` or by using a 1x1 px transparent replacement image) is "nicer" to the flow than a `display:none`.  
Probably it's beneficial to have an option where the user can choose between reserving the space (i.e. `visibility: hidden`) and not preserving the space (the `display:none`).

By "map should work" I'm thinking of defining a mapping between e.g. integer values and URL strings. For static maps you could e.g. use something like
```
    <mappings>
        <mapping name="meteograms">
            <entry value="0">https://www.yr.no/place/Germany/Bavaria/Holzkirchen~2899676/avansert_meteogram.png</entry>
            <entry value="1">https://www.yr.no/place/Germany/Bavaria/Munich~2867714/avansert_meteogram.png</entry>
        </mapping>
    </mappings>
``` 
Then you could change the displayed meteogram by a simple KNX message between two different places (e.g. between home, work and holiday home).

peuter:
I just added a `placeholder` attribute to let the user decide how to handle the "no image url given" case.

I think the mapping feature should be part of another PR, if someone needs it (i don't). For openHAB backends there is just no need for that at all, because you can directly send the image URL and that is way more flexible than the mapping approach.

ChristianMayer (approved):

PR #1051, opened 2021-03-02
Fix code that caused an infinite loop of `__makeNavbarValid` calls

ChristianMayer:
Fix code that caused an infinite loop of `__makeNavbarValid` calls when the top and/or the bottom navbar were not used (i.e. in all practical cases)

### DIFF
--- a/source/class/cv/ui/layout/ResizeHandler.js
+++ b/source/class/cv/ui/layout/ResizeHandler.js
@@ -225,25 +225,6 @@ qx.Class.define('cv.ui.layout.ResizeHandler', {
     },
 
     __makeNavbarValid: function() {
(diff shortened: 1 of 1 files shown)


ChristianMayer:
@peuter just throwing some code out fixed the issue for me - and I couldn't find any negative effects during my trials. But I wasn't around that part of the code since ages so I really hope I didn't break anything anywhere else.

peuter:
I have no idea what this code was good for, because I don' t see any relevance for top/bottom navbar for column width calculation. Maybe a leftover of some old code refactoring and obsolete now. I also did not find any differences with this fix.

peuter (approved):

PR #1052, opened 2021-03-13
Backend: client for MQTT

ChristianMayer:
This PR allows the CometVisu to connect to a MQTT broker using websockets.

All the (basic) functionality should work. But a second PR with some refinement, documentation, ... will come at a later stage.

This PR also changes the internal address structure from an array with fixed indices to a hash/object to help with code readability.


source/class/cv/io/mqtt/Client.js
```diff
@@ -0,0 +1,263 @@
...
+    getProviderConvertFunction : function (name) { return null; },
+
+    /**
+     * Set a subset of addresses the client should request initially (e.g. the ones one the start page).
+     * This can be used to increase the init state loading speed by sending an initial request with a smaller
+     * subset of addresses to the backend and send the rest later.
+     * @param addresses {Array}
+     */
+    setInitialAddresses: function(addresses) {
+      console.log('subscribe Initial: ', addresses);
```
peuter, line 101: I guess this can be removed

source/class/cv/io/mqtt/Client.js
```diff
@@ -0,0 +1,263 @@
...
+
+      if( null !== credentials && 'username' in credentials && null !== credentials.username ) {
+        options.userName = credentials.username;
+      }
+      if( null !== credentials && 'password' in credentials && null !== credentials.password ) {
+        options.password = credentials.password;
+      }
+
+      try {
+        this.client = new Paho.MQTT.Client(this._backendUrl, 'CometVisu_' + Math.random().toString(16).substr(2, 8));
```
peuter, line 163: Please define this variable at the beginning if the members section. It would also be better to use and underscore for this to indicate that this this an internal (protected) variable: e,g,

```
members: {
   /*
   * @var {Paho.MQTT.Client}
   */
  _client: null,
  
  ....
  
  this._client = ...

File: source/class/cv/io/mqtt/Client.js [hunk 3]
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * Can be overridden to provide an error handler for client errors
+     * @param type {Number} one of cv.io.Client.ERROR_CODES
+     * @param message {String} detailed error message
+     * @param args
+     */
+    showError: function(type, message, args) {}
+  }
+
```
peuter, line 262: Please add a constructor to clear local objects (at least the _client):
```js
 destruct: function () {
   this,_client = null;
 }
 ```
 
 Although the backend client is a singleton that will not be destroyed during the livetime of the current cometvisu instance, we should "just do this" to be clean.

peuter (changes requested):

peuter (approved):

PR #1057, opened 2021-04-03
New tree editor for schema based XML-files

peuter:
replacement for the old visual editor. Complete rewrite with better integration into the new manager and some improvements compared to the old editor, like:
* support for comments
* support to re-order elements via drag&drop
* touchscreen friendly


source/class/cv/ui/manager/dialog/ValidationError.js
```diff
@@ -0,0 +1,174 @@
...
+     * Create the main content of the widget
+     */
+    _createWidgetContent: function() {
+      let container = this._createDialogContainer();
+      let vbox = new qx.ui.container.Composite(new qx.ui.layout.VBox(10));
+      container.add(vbox);
+
+      let prologue = this.tr('<h3>Config Errors</h3>\
+<p>The validation of "%1" showed some errors. It is NOT recommended to edit this file in the XML-Tree editor, because \
+if can break the file completely. Please open this file in the text editor and fix the errors there. \
```
ChristianMayer, line 76: "because if can break" -> "because it can break"

source/class/cv/ui/manager/dialog/ValidationError.js
```diff
@@ -0,0 +1,174 @@
...
+Only proceed to edit the file in the XML-Tree editor if you know what you are doing.</p>', this._file.getFullPath());
+
+      let options = this.tr('<p>You can choose between the following options:\
+<ul>\
+    <li><b>Cancel</b> - Close this editor</li>\
```
ChristianMayer, line 81: Cancel vs. Close vs. the button is labeled "Close"

ChristianMayer:
In the manager, clicking the `+`:
grafik
created a new, empty config in the text editor. (That's sort of expected).
After saving it without any further action it appears in the configs list, but I can't open it with the new editor, as the icon is missing:
grafik
And the context menu is greyed out:
grafik

ChristianMayer:
In the editor dragging the big "(+)" to create a new entry in the plugins section shows a console error:
```
Native.js:48 336597 qx.ui.core.queue.Appearance: Error in the 'Appearance' queue:TypeError: Cannot read property 'height' of null TypeError: Cannot read property 'height' of null
https://timberwolf76.local/proxy/visugit/obiwan/cv/package-53.js:2963:71,https://timberwolf76.local/proxy/visugit/obiwan/cv/package-39.js:1821:37,) [as series] (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:21759:24,https://timberwolf76.local/proxy/visugit/obiwan/cv/package-39.js:1819:33,) [as then] (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:21600:39,) [as _dispatchEvent] (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-39.js:1818:24,https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:18356:39,) [as then] (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:21600:39,ht…erwolf76.local/proxy/visugit/obiwan/cv/package-1.js:18355:28,Array.forEach (<anonymous>),) (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:18343:21,https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:3930:31,) [as dispatchEvent] (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:12769:52,https://timberwolf76.local/proxy/visugit/obiwan/cv/package-7.js:2053:46,) [as then] (https://timberwolf76.local/proxy/visugit/obiwan/cv/package-1.js:21600:39,https://timberwolf76.local/proxy/visugit/obiwan/cv/package-7.js:2042:26
```

Dragging it between `plugins` and the first `plugin` entry most times let me choose `#comment` and `plugin` (as expected) and sometimes it shows me the full list that might be inserted below `meta`:
grafik

When `plugin` appears and I add a new plugin entry it doesn't appear in the tree.

Dragging the `(+)` not between the lines (i.e. a displayed line will appear) but on the parent entry still shows a `+` on the mouse courser although releasing the mouse button just silently cancels the action. IMHO it would be better when that action would insert a new child instead (e.g. on the first position as it might be that the user was aiming for the line but just missed it by a slight movement)

ChristianMayer:
Trying to add a new entry at the empty config failed in all different ways I have tried. The different ways showed different messages on the console:
grafik

source/class/cv/ui/manager/dialog/ValidationError.js
peuter: Fixed

source/class/cv/ui/manager/dialog/ValidationError.js
peuter: Renamed the button

peuter:
Its hard to refer to a single error report this way. I think I fixed all of the reported ones, despite from the console error `qx.ui.core.queue.Appearance: Error in the 'Appearance' queue:TypeError:` which I could not reproduce.

As an explanation for creating new element via drag&drop: The indicator line should show where the new element will be added. When you are dragging between an Element and its first child its hard to distinguish, but the indicator line show by its starting point on the left if you are adding a new "first-child" or a new sibling of the parent node. I know its far from clear and no ideal solution but I didn't find a better one yet.

ChristianMayer:
The Appearance bug is constantly happening for me. Digging deeper is complicated to impossible as no source map is found. The best I could do was to set a break point at Native.js:48 and have a look at the Call Stack.
But the calls are not accessible to peek inside, they look like:
grafik
Pretty print doesn't help, it doesn't really change a thing.

Exactly the same happens with firefox as well: the same error message after the same action (dragging the `(+)` to `plugins`) and the same inaccessability of the stack trace

ChristianMayer:
Editing an address element the transform stays empty and throws (especially during scrolling) many of those:
grafik

ChristianMayer:
I can't find a way to enter an address. Neither by `(+)` nor by right click it's possible for me to enter it:
grafik

peuter:
Adding text-nodes as child is possible now (fixes the missing address value entry). The generation of the available Transforms is also fixed.

I still cannot reproduce the console error: Tried source and build versions in chrome and firefox browsers, there is no error message in the console when I drag from + to plugins.

peuter:
The console error that appeared when opening the a dialog has been fixed in the qxl.dialog package, I upgraded the required version of that package. The easiest/safest way to upgrade might be by just deleting the qx_packages folder and let the compiler re-build it (is done automatically before compiling).
I also fixes some errors I found. I am pretty sure that there are still errors hidden inside this PR but from my point-of-view its ready to be merged

ChristianMayer:
The Editor does add multiple addresses to the `<address>` element (wrong and bad user experience) and thus creates a config that's not containing a valid address:
grafik

ChristianMayer (commented):
- [ ]  When the `(+)` is above tree lines (due to large tree / small screen) the entries can not be selected by mouse:

grafik

ChristianMayer (commented):
- [x] The `#text` input dialog is too small for a bit longer texts:
grafik

(Possible solution: leave it mostly as it is, but offer an "..." button that changes from line input to a text area)

ChristianMayer (commented):
Adding a `influx` child to the `diagram` doesn't work and shows on the console:
grafik

ChristianMayer (commented):
- [x] Adding an `address` doesn't show the available ones with their description and the corresponding transform.
(I.e. the answer from list_all_addresses.php from the old editor)

peuter:
Everything, despite from the influxdb stuff should be fixed now.

ChristianMayer:
Dragging the `(+)` in the tree at one (non specific) point showed in the console:
```
Tree.js:812 Uncaught TypeError: Cannot read property 'getContentLocation' of null
    at wrapper.onDrag (Tree.js:812)
    at EventHandler.js:316
    at Function.qx.event.Utils.series() [as series] (Utils.js:332)
    at EventHandler.js:314
    at Function.qx.event.Utils.then() [as then] (Utils.js:159)
    at wrapper.qx.ui.core.EventHandler.prototype._dispatchEvent() [as _dispatchEvent] (EventHandler.js:313)
    at AbstractBubbling.js:238
    at Function.qx.event.Utils.series() [as series] (Utils.js:332)
    at AbstractBubbling.js:223
    at Function.qx.event.Utils.series() [as series] (Utils.js:332)
    at AbstractBubbling.js:220
    at Function.qx.event.Utils.then() [as then] (Utils.js:159)
    at wrapper.qx.event.dispatch.AbstractBubbling.prototype.dispatchEvent() [as dispatchEvent] (AbstractBubbling.js:215)
    at wrapper.qx.event.Manager.prototype.dispatchEvent() [as dispatchEvent] (Manager.js:975)
    at Object.qx.event.Registration.dispatchEvent() [as dispatchEvent] (Registration.js:272)
    at wrapper.qx.event.handler.DragDrop.prototype.__fireEvent__P_213_17() [as __fireEvent__P_213_17] (DragDrop.js:488)
    ```

**Comment by ChristianMayer** [comment]
Adding multiple `#text` to a `<label>` is still possible, but it's not working properly. Sometimes the additional `#text` does appear and sometime not. And the export to the confix-xml doesn't create an expected result as the first element is added last:
grafik

**Comment by ChristianMayer** [comment]
Elements with no attributes should show in the edit dialog that there's nothing to edit to prevent user confusion:
grafik
Suggestions:
1. Change the title to `Bearbeite "label" Attribute`
2. Show Text `Element "label" hat keine Attribute zum Editieren` in the middle of the dialog box
3. (Optional, when simply doable): Show only the "Ok" Button and hide "Abbruch" in the case of no attributes

**Comment by ChristianMayer** [comment]
Mobile (on a tablet) doesn't work completely:
- Quickly dragging `(+)` doesn't do anything
- Tap on `(+)` and wait a bit create a shot shake and I can drag a `[+]` on the screen. The `[+]` is hardly visible as it's hidden by the finger doing the touch. And dragging it over the tree doesn't do anything
- Tap and holing an element in the tree shows the context menu (like right mouse click) but it disappears after ~1 second again, i.e. I don't have enough time to use it
- Tab and wait a little bit on a tree entry created a copy of it in the tree, even in places where it doesn't belong (like `page` above `meta`). These are also "invalid" copies as changeing the name of one of those entries renames all of them:
   Screenshot_20210410-224047_Chrome

Additional usability improvement (for desktop and mobile): Single click on `(+)` should not show "Bitte ziehen sie diesen Button auf die Stelle an der das neue Element platzert werden soll." but instead work as choosing "+ Kind-Element einfügen" from the selected line.

**Comment by ChristianMayer** [comment]
Opening the help shows a strange screen (I guess the editor overlay get's made too small so that the underlaying original config is shown):
grafik

Also, please recheck the German help text as there are a few grammar errors

**Comment by peuter** [comment]
> Dragging the `(+)` in the tree at one (non specific) point showed in the console:

Where is the non specific point? I can't reproduce that.

> Adding multiple #text to a <label> is still possible...

Thats actually correct, label has a mixed content and you can have e.g. a child sequence of `#text`, `icon`, `#text`. So we have to allow multiple text children. The errors with positioning the additional text child will be fixed.

> Elements with no attributes should show in the edit dialog that there's nothing to edit to prevent user confusion

The current implementation is to not open the dialog when there is nothing to edit. There's just a bug with elements with mixed content like the label. I will fix that.

**Comment by peuter** [comment]
> Quickly dragging (+) doesn't do anything

Thats actually on purpose, because dragging on touchscreen starts with a delay to distinguish between a drag and a scroll event. Quickly dragging is the scroll gesture on touch devices.

The other touchscreen and drag / contextmenu problems should be better handled / fixed now.

**Comment by ChristianMayer** [comment]
Dragging the `(+)` to insert a new element to the bottom of a page (I guess: the last page) doesn't work although it's showing the insert line:
grafik

I guess it's a conflict about interpretation whether the user wants to add something to the end of the topmost element (`pages` in this case) or the bottom most one (`line` in this case).
My intuitive interpretation would be to add something as the same level of the last shown object, i.e. a `page "Start"` child after the `line`. This interpretation is confirmed by the start point / length of the shown insert line.

**Comment by ChristianMayer** [comment]
> > Dragging the `(+)` in the tree at one (non specific) point showed in the console:
> 
> Where is the non specific point? I can't reproduce that.

That's why I wrote "(non specific) point" as I also can't reproduce it. It just has happened by playing around with it - and so I fear it will also happen to other users in the future.

**Comment by ChristianMayer** [comment]
At `text/label/#text` element:
Right click "copy" and at a different `text/label/#text` a "paste" does nothing.

Right clicking and selecting "copy" on a node doesn't make "paste" appear in a different node that should be able to take a paste (e.g. a whole widget and a page or copying a `layout` from one widget to an other one)

**Comment by ChristianMayer** [comment]
- [ ] The highlighting breaks when the tree gets modified structurally (elements moved around, elements created or deleted)

When this can't be fixed easily (as this would require some sort of tracking) I could imagine a solution where after a "critical action" the highlighting will be disabled and the refresh button will be highlighted instead. Probably together with a tooltip that appears for a short time and says "Please refresh to activate highlighting again".

**Review by ChristianMayer — CHANGES_REQUESTED** [review]
- [x] There's some bigger issues with config file generation:

1. `text/label/x` existed already (can't remember whether it was shown in the preview though)
2. Adding `aaa` and `zzz` childs worked fine. They are also shown in the preview - but not the `x`?!?
3. Adding `middle` also worked, but still no `x` in the preview
4. Changing `x` to `xxx` still only shows `aaamittezzz` in the preview. Changing it back to `x` doesn't change the preview
5. The last `text/label/Welcome...` is never shown in the preview, highlight works though
6. Adding a new `text` by right click on the `page "Start"*` creates the node at the top. (I'd expect to be at the bottom, but that's no big issue)
7. Dragging the `::` of the new node to the very bottom doesn't do anything 
8. Dragging the `::` of the new node to just above the last text node is deleting it
9. 
grafik

**Comment by ChristianMayer** [comment]
If possible: for mobile devices the spacing of the elements should be a bit increased to take account for "fat fingers"

**Comment by peuter** [comment]
> At text/label/#text element: Right click "copy" and at a different text/label/#text a "paste" does nothing.

see below

> There's some bigger issues with config file generation

Please re-test with the current fixes. Main problem was that the XML-Node has not been cloned on copy & paste, and you had two different elements in the tree that acually worked on the same node. That lead to lots of follow up errors.
I also tried to improve the handling of text nodes in mixed content (e.g. label). Having two text-nodes as direct sibling leads to unpredictable behaviour (the parser might parse them as one node on the next parsing). I added a constraint that should only allow childrens like `#text' , `icon`, `#text` but not `#text`, `#text`. That constraint is not covered by the schema, but if we do not do that we have strange behavior (its basically the same problem in the old editor)

> Dragging the (+) to insert a new element to the bottom of a page (I guess: the last page) doesn't work although it's showing the insert line

Should be fixed now.

> If possible: for mobile devices the spacing of the elements should be a bit increased to take account for "fat fingers"

I already tried that, but every attemp to make it bigger failed so far. There must be some place where the height is overridden in the tree widget, I wasn't able to find a way to manage that. I will first focus on bugs and address these UX improvements when the bugs are fixed.

**Comment by peuter** [comment]
> Dragging the (+) in the tree at one (non specific) point showed in the console

That one should be fixed too, because of
```js
if (!orig) {
  return;
}
...  
const origCoords = orig.getContentLocation();
```
and that is the only place where getContentLocation is used.

peuter:
> The highlighting breaks when the tree gets modified structurally (elements moved around, elements created or deleted)

There are three modes detected now and the current state is shown above the preview:
1. Preview is in sync: Refresh-preview button is disabled, everything is fine, highlighting is enabled
2. There have only been content changes (attribute values / text content): Refresh-button is enabled, highlighting is enabled, text states that the preview does not show the current state and can be refreshed
3. There is at least one structural change (child deleted, moved, added): refresh-button is enabled, highlighting is disabled, text states that the highlighting is disabled until the preview is refreshed

ChristianMayer:
- [ ] Fat finger size is much better now. But text dialog might need a little bit of tweaking as shown here on a Galaxy Tab:
tempFileForShare_20210413-223303

ChristianMayer (changes requested):
- [ ] Pulling `infotrigger` above `info` does work. Pulling `info` below `infotrigger` does not work, even when hitting the few pixels where the line is appearing. (At least most of the time.)
grafik

ChristianMayer (changes requested):
Pulling `(+)` to insert a `label` or a `layout` to the info doesn't show those valid elements, only `address` and `#comment`:
grafik

ChristianMayer (changes requested):
- [ ] Resorting by drag is off by one: The `infotrigger`  was dragged between `text` and `info` but it appears on the bottom
grafik

ChristianMayer:
Optimization: Clicking this "unmodified" shouldn't reload as it might come unexpected and take quite a while:
grafik

ChristianMayer (changes requested):
- [ ] On a tablet: The burger menu button opens the context menu, but it doesn't react to any tab. Also the burger buttons are stopping working then
- [ ] Adding the `colorchooser` doensn't work, console says 
```
Tree.js:1216 Uncaught (in promise) TypeError: Cannot read property 'then' of undefined
    at Tree.js:1216
```

peuter:
> Adding the colorchooser doensn't work, console says...

Problem with adding Widgets that do not have attributes to edit is fixed.

> Optimization: Clicking this "unmodified" shouldn't reload as it might come unexpected and take quite a while:

Fixed

> Pulling (+) to insert a label or a layout to the info doesn't show those valid elements, only address and #comment:

Position checking when adding / moving elements in sequences (e.g. most widgets have the sequence: "layout", "label", "address" has been improved. My several tests went fine, but there might be other corner cases I did not think of.

> Pulling infotrigger above info does work. Pulling info below infotrigger does not work, even when hitting the few pixels where the line is appearing. (At least most of the time.)

Moving / creating element below the last element in the tree is allowed now (max 20px below)

The other reported errors are still on my todo list. Until I'm done please stop testing (or just test the fixed ones).

> Resorting by drag is off by one: The infotrigger was dragged between text and info but it appears on the bottom

I wasn't able to re-produce that, either that problem has been fixed by the other fixes or I need more info on this (dragged from where, did you move it or created it?)

peuter:
> Fat finger size is much better now. But text dialog might need a little bit of tweaking as shown here on a Galaxy Tab:

That actually requires a bugfix in the TextArea component of qooxdoo. I will create the PR for that later, we should not block this PR by that.

peuter:
Ok I'm done with my todo-list now. Feel free to test again.

peuter:
OK just found another bug:

- [ ] Moving an element down on the same level inserts it a a wrong position (1 below the correct one)

ChristianMayer (changes requested):
- [x] Dragging the `(+)` below the last entry when it's childs are visible does open the attribute edit dialog and not the dialog to be able to insert a new element:
grafik

ChristianMayer (changes requested):
- [ ] Not configured InfluxDB connection fails with (only) console error messages:
grafik
grafik

ChristianMayer (changes requested):
- [ ] On a mobile device (tablet) the tree triangles to show/hide the childs has a too small active area, i.e. it's too hard to hit it with the finger

ChristianMayer (changes requested):
- [ ] Dragging the three widgets to resort them does sometimes work and sometimes not, although  I think I'm doing always the same. Draging upwards seems to work more often than dragging an element to the end. And when the screen is so small that the `(+)` is above the tree it's quite often (but not always!) that dragging something to the end doesn't work:
grafik

ChristianMayer (changes requested):
- [ ] Dragging the `colorchooser` to the end (probably touching the `(+)`) shows on the console:
grafik

peuter:
> Dragging the (+) below the last entry when it's childs are visible does open the attribute edit dialog and not the dialog to be able to insert a new element:

Well, the only allowed element left to be added as a child into this info-element is another address-element. And the selection dialog is not shown when there is nothing to select from. So this is actually a feature not a bug.

peuter:
> Not configured InfluxDB connection fails with (only) console error messages

The problem that the edit dialog did not open is fixed. Also there is a message shown when the data-request failed (at least is the user is not able to enter data manually because he sees an empty SelectBox in that case and needs to know whats going on). For ComboxBoxes (when the user can still enter custom values) the message is shown as a placeholder (note that the placehlder is only shown if the value eis empty).

> On a mobile device (tablet) the tree triangles to show/hide the childs has a too small active area, i.e. it's too hard to hit it with the finger

Area is bigger now

> Dragging the three widgets to resort them does sometimes work and sometimes not, although I think I'm doing always the same. Draging upwards seems to work more often than dragging an element to the end. And when the screen is so small that the (+) is above the tree it's quite often (but not always!) that dragging something to the end doesn't work:

I moved the + Button out of the way during dragging, please check if that helps. If not please check if there is any log message in the console about (something like "NOT ALLOWED add info after layout").

> Dragging the colorchooser to the end (probably touching the (+)) shows on the console

Please also check if moving the +-button away helped here. Otherwise I need to know what "target" is in this case.

ChristianMayer:
> > Dragging the (+) below the last entry when it's childs are visible does open the attribute edit dialog and not the dialog to be able to insert a new element:
> 
> Well, the only allowed element left to be added as a child into this info-element is another address-element. And the selection dialog is not shown when there is nothing to select from. So this is actually a feature not a bug.

I'd expect that it behaves consistently. Dragging the `(+)` somewhere never opens the attribute dialog box - except in the case I had described.

ChristianMayer (changes requested):
- [ ] The triangles are now broken on the desktop: 
grafik

peuter:
> > > Dragging the (+) below the last entry when it's childs are visible does open the attribute edit dialog and not the dialog to be able to insert a new element:
> > 
> > 
> > Well, the only allowed element left to be added as a child into this info-element is another address-element. And the selection dialog is not shown when there is nothing to select from. So this is actually a feature not a bug.
> 
> I'd expect that it behaves consistently. Dragging the `(+)` somewhere never opens the attribute dialog box - except in the case I had described.

It always opens the dialog box, after you have made the selection on what to add. And if you do not have to make a decision about which type of element to add herem because only one is allowed we can skip this step.
I don't see any sense to show a selection dialog when the user has nothing to decide just for the sake of consistency.

peuter:
> The triangles are now broken on the desktop:

Fixed

peuter:
Screenshots in documentation are up-to-date now and I did not find an error during my last tests. Feel free to start testing again.

ChristianMayer (changes requested):
- [x] Selecting an element in the tree by simple left click doesn't work anymore.
Only a right click does work.

peuter:
> Selecting an element in the tree by simple left click doesn't work anymore.
> Only a right click does work.

Works perfectly for me. Do you see any errors on the console? Which Browser / Device (touchscreen or not)? Working fine on Chrome for me. I need more details to be able to reproduce that.

ChristianMayer (changes requested):
- [ ] Dragging the `(+)` is, especially on the last element, not behaving predictable:
 - just dragging it upwards does nothing 
grafik
 - Dragging it a bit higher shows the line but nothing happens when I release the button: 
grafik
 - Not releasing the button but moving the drag further down shows a line where none was before (releasing the button still does nothing, although the mouse icon has changed) 
grafik
 - and moving it even further down still shows the same behavior: 
grafik

The line and the mouse cursor `[+]` should only appear when something would happen there when the pressed mouse button is released

ChristianMayer:
- [ ] adding an audio widget opens a too small window: 
grafik

peuter:
> Dragging the (+) is, especially on the last element, not behaving predictable:

Well you have to drag it until you reach an item in the tree to be able to have a target and check if we can drop it there. Or more technically: I need at least one "dragover" event to start checking and the free area does not trigger that event. When you leave the last item and drag into the free area again, its the other way around, no new "dragover" events is fired and therefore that last check is still valid (if an item can be added after the last item). I am not sure that I can find a workaround for this problem.

ChristianMayer:
After the reload I can't see the `[+]` at the mouse cursor anymore, only the not-allowed-sign. But releasing the mouse button here does open the dialog: 
grafik

ChristianMayer:
- [ ] having diagram/audio/audio and resort-dragging diagram between the audios did work. But dragging one of those to the very bottom (neither as line nor in the big empty space below) did not work. And quite often the next dragging also doesn't work.
  Actually the dragging does quite often not work. But on the second attempt exactly the same drag does work. Messages in the console are `dropping not accepted here`, `not allowed as child element of audio`  and `dropping on same element forbidden`
grafik

ChristianMayer:
- [ ] dragging the `(+)` upwards to add a new widget and waiting for the line to show does quite often nothing on release of the mouse button. The Console says `drop target not found`

ChristianMayer:
- [ ] sporadic issue (happend only one): after adding the colorchooser widget I get those messages irrelevant for the user on the main screen: 
grafik

peuter:
>  sporadic issue (happend only one): after adding the colorchooser widget I get those messages irrelevant for the user on the main screen:

Those messages are shown as a feedback after you have moved an item, because sometimes its hard to see that the item really has been moved, especially when you have moved items of the same type. These snackbar messages should improve the user experience by giving him a hint that something has really happened. But they should not occur in such a bulk (unless you have moved lots of items within some seconds). I have never seen that behaviour. Maybe its just a follow-up error of another one.
So let me just fix the other errors and than you can check if this still happens.

ChristianMayer:
- [ ] Dragging the `(+)` to the `diagram` let's me add an `rrd` child. Just clicking `ok` in the appearing dialog does nothing, i.e. no `rrd` child appears in the tree
grafik

ChristianMayer:
- [ ] a resort drag below the last element doesn't work when it's childs are visible (here: `diagram` can't be dragged below `audio`. But with collapsed `audio` it's possible)
grafik

peuter:
> a resort drag below the last element doesn't work when it's childs are visible (here: diagram can't be dragged below audio. But with collapsed audio it's possible)

It not a bug its a feature ;-) Jokes aside its nearly impossible to decide if the user wants to drag below the child or the parent. So I made the decision that you always target the last shown element. If you want to drag below the parent you have to collapse it. That is a quite clear rule the user has to learn.

peuter:
> At the same setup: dragging the (+) to show the line above the influx child opens the new element dialog, as expected. But dragging it to show a line below the influx shows the attribute edit dialog of the influx child. Expected behavior would be to also show the new element dialog

That is the defined order in the schema. 
```xml
<xsd:sequence>
      <xsd:element name="layout" type="layout" minOccurs="0" maxOccurs="1"/>
      <xsd:element name="label" type="label" minOccurs="0" maxOccurs="1"/>
      <xsd:element name="axis" type="axis" minOccurs="0" maxOccurs="unbounded"/>
      <xsd:element name="rrd" type="rrd" minOccurs="0" maxOccurs="unbounded"/>
      <xsd:element name="influx" type="influx" minOccurs="0" maxOccurs="unbounded"/>
  </xsd:sequence>
  ```
  So if you have an existing `influx` child, after that one only more `influx` childs are allowed. Otherwise the config would be invalid.

ChristianMayer:
In that case the element add dialog box should come and let me add a new `influx` element. (Or add it directly).

But I'd never expect that the attribute edit dialog opens up by dragging the `(+)`

peuter:
> In that case the element add dialog box should come and let me add a new `influx` element. (Or add it directly).
> But I'd never expect that the attribute edit dialog opens up by dragging the `(+)`

I doubt that it does that (but I havent checked myself). It opens the edit dialog for the new influx element, as it always does when only one possible element is allowed at that position. The dialog should indicate in the title that a new influx element is created, but maybe we have to make that situation more clear.

peuter:
> The dialog should indicate in the title that a new influx element is created, but maybe we have to make that situation more clear.

Ok that is indeed missing completely, I'll add that.

ChristianMayer:
https://user-images.githubusercontent.com/10218721/116788800-ff0b6b00-aaab-11eb-8469-b036992051b6.mp4

ChristianMayer:
- [ ] probably got lost in the mean time: https://github.com/CometVisu/CometVisu/pull/1057#issuecomment-819035567

ChristianMayer:
- [ ] adding a `text` widget and then a `#text` node below it's `label` it can't be edited / changed. Opening an already existing `text` widget with content the text content can be edited though.

peuter:
> Dragging the (+) to the diagram let's me add an rrd child. Just clicking ok in the appearing dialog does nothing, i.e. no rrd child appears in the tree

That is fixed now. And it should by clear wether you edit an existing element or a new one. The dialog has a title and an additional message with that information

peuter:
> adding a text widget and then a #text node below it's label it can't be edited / changed. Opening an already existing text widget with content the text content can be edited though.

Fixed

> probably got lost in the mean time: #1057 (comment)

No, works for me on android tablet and ipad. Is this still an issue?

peuter:
I think I fixed or explained all new issues now. Please verify.

ChristianMayer:
Point 1 of https://github.com/CometVisu/CometVisu/pull/1057#issuecomment-830651491still doesn't work

peuter:
This is the only text in the whole linked comment:

> Fat finger size is much better now. But text dialog might need a little bit of tweaking as shown here on a Galaxy Tab:

So what do you mean with point 1?

peuter:
> Point 1 of [#1057 (comment)](https://github.com/CometVisu/CometVisu/pull/1057#issuecomment-830651491still) doesn't work

Wait, the link is not working, maybe you meant another comment?

peuter:
I guess its about this one:

> Dragging the (+) to the diagram let's me add an rrd child. Just clicking ok in the appearing dialog does nothing, i.e. no rrd child appears in the tree

BUt I fixed that one, do you see errors in the console when you do that?

peuter:
Ok found it, it does work when you drop close to the upper edge if the influx element but it does not work when you drop close to the lower edge of the diagram element.

peuter:
Should work on both positions now.

ChristianMayer:
All the issues raised so far seem to be fixed, except the text dialog box on my tablet (see also Gitter). :+1: 

From further testing I found this minor issue: A dark blue link is quite hard to read on a dark background: 
grafik

peuter:
Right, I changed the link color to one with a better contrast

ChristianMayer:
- [ ] having multiple pages: selecting a widget on a parent-page switches pages and shows the highlight. But selecting a widget on a child page doesn't switch pages to the child page

ChristianMayer:
- [ ] there's constantly a timer running, even when doing nothing. This can drain the battery.

ChristianMayer:
- [ ] Saving destroyed the links in the footer: 
grafik

ChristianMayer:
- [ ] The "invalid config" message can become too bit to fit on the screen and thus the buttons can't be pressed:
grafik

ChristianMayer:
- [ ] pressing "Fortfahren" doesn't close the dialog.

Just afterwards this error shows on the console:
grafik

The config is loaded in the background. But as the dialog doesn't go away that doesn't change much for the user.

peuter:
> there's constantly a timer running, even when doing nothing. This can drain the battery.

I don't what you main, what timer is running where?

peuter:
> Saving destroyed the links in the footer:

Fixed

> pressing "Fortfahren" doesn't close the dialog.

Errors are catched now, although its unpredictable what happens when you load an invalid element. But the hint in the dialog should have made that clear.

Also the creation of invalid elements by not setting a value to a required attribute should be fixed now too.

peuter:
> The "invalid config" message can become too bit to fit on the screen and thus the buttons can't be pressed:

That is fixed too, but there still is a limit of a minimum height where the button will not be visible. When someone uses the manager or this editor with a screen height below round about 400px I cant help him.

ChristianMayer:
- [ ] (Forced) Opening of the malformed config where the escaping of the statusbar got missing (-> https://github.com/CometVisu/CometVisu/pull/1057#issuecomment-830697690 ) destroys the tree when one of the status elements is clicked on:
grafik

Expected behavior when a malformed config is force opened: the editor should behave graceful and probably limit its functions but it shouldn't be corrupted.

ChristianMayer:
- [ ] inserting a space character doesn't work here, normal letters do work. Backspace does work. And sometimes (I guess after inserting a normal letter and then pressing backspace) a space does work.
grafik

peuter:
> (Forced) Opening of the malformed config where the escaping of the statusbar got missing (-> #1057 (comment) ) destroys the tree when one of the status elements is clicked on:

This case is fixed, but I dont thinks it is possible to harden an schema based xml editor that much that it can handle every broken xml code

> inserting a space character doesn't work here, normal letters do work. Backspace does work. And sometimes (I guess after inserting a normal letter and then pressing backspace) a space does work.

I tried all combination I could think of but never ran into that problem (new empty node, eiting an existing one, deleting existing content and editing then) => in all cases I could enter every possible key. Are there any errors on the console when you do that.

ChristianMayer:
> > (Forced) Opening of the malformed config where the escaping of the statusbar got missing (-> #1057 (comment) ) destroys the tree when one of the status elements is clicked on:
> 
> This case is fixed, but I dont thinks it is possible to harden an schema based xml editor that much that it can handle every broken xml code

I *guess* the problem is that the tree tries to show the text content, but that content contains a HTML element. In that case it should be ensured that that the user/config settable text doesn't get displayed as an HTML element.
 
> > inserting a space character doesn't work here, normal letters do work. Backspace does work. And sometimes (I guess after inserting a normal letter and then pressing backspace) a space does work.
> 
> I tried all combination I could think of but never ran into that problem (new empty node, eiting an existing one, deleting existing content and editing then) => in all cases I could enter every possible key. Are there any errors on the console when you do that.

I can reproduce it by opening the (empty) default config and then going there to one of the statusbar/status elements. Probably because they use `cdata`. Some do work and some don't.
Sometimes it does work on a multline entry (e.g. the first one with `<img...`), but then clicking on a different line it doesn't work any more.
The console doesn't show any message.

peuter:
> I guess the problem is that the tree tries to show the text content, but that content contains a HTML element. In that case it should be ensured that that the user/config settable text doesn't get displayed as an HTML element.

Not really the problem was, that while writing the editor i assumed that it is not possible to create or edit an invalid config with this editor. So when an element loaded its children and one of them was not allowed there i threw an exception to make sure that this case was noticed. That exception broke the rendering process because it was not captured anywhere. I changed that to only mark the element as invalid and skip loading the invalid child. In these cases you are not able to fix the config with this editor, because the "problem" is not loaded. But this editor is not meant to be able to fix a broken config, thats what the text editor is for. You can fix wrong attribute or text values, but not structural problems like this one (having a child element where only text is allowed, or having an child that is not allowed there).

peuter:
I can reproduce the missing whitespaces now. I happens without a real pattern, sometimes immediately sometime, after I did some editing. The problem is inside the monaco editor that is used there. I tried every space-related setting I found, but none of them fixed it. I also did not find a bug report in the monaco editor for that problem. For the moment I don' t know how I can fix the problem.

ChristianMayer:
Is the monaco editor needed here?
The user will edit only a very small snippet of text and doesn't really need syntax highlighting here, so I guess a simple textarea should work fine.

peuter:
Its not needed, but helpful. Especially for users that are nor experiences in writing html code. I would suggest to disable it for now and open a separate issue as a reminder and try wo solve it later.

peuter:
Fancy editor is disabled, the TextArea used now works as expected.

ChristianMayer (approved):

PR #1064, opened 2021-04-24
use indexedDB instead of localStorage

peuter:
because it has no such low limits for quota. This avoids broken cache due too quota exceeded exceptions.

### DIFF
--- a/source/class/cv/Application.js
+++ b/source/class/cv/Application.js
@@ -179,6 +179,7 @@ qx.Class.define("cv.Application",
      * during startup of the application
      */
     main : function() {
+      cv.ConfigCache.init();
       qx.event.GlobalError.setErrorHandler(this.__globalErrorHandler, this);
       if (qx.core.Environment.get("qx.debug")) {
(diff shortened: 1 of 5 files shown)


ChristianMayer:
Trying the code (Chrome, HTTPS connection) I can see that IndexDB has one object storage.

But looking at it it seems that nothing gets stored: "Total entries: 0"
grafik

peuter:
Please keep in mind, that caching is disabled in source builds. You have to explicitly enable it with enableCache=true

ChristianMayer:
Thanks for the hint, that was the reason.

But, different issue, with URL `.../?enableCache=true#id_`:
grafik

peuter:
Ah right, I did not test with the default config (with no name). The name is used as key in the DB. That obviously does not work when there is no name. Will fix that soon. But it should work with configs with names.

ChristianMayer:
Yes, config with not-default-name worked fine

peuter:
default should work now too

ChristianMayer (approved):

PR #1065, opened 2021-04-24
Change regex for incapable safari.

### LINKED ISSUE #1063: Safari does not support lookbehind regex used in mqtt transform
`/(?<!\\)\./` breaks on safari and causes a loading error on safari browsers (you only see a white page): `Invalid regular expression: invalid group specifier name`

See:
https://caniuse.com/js-regexp-lookbehind

That is the cause of the error reported here: https://knx-user-forum.de/forum/supportforen/cometvisu/1646571-nutzt-jemand-0-12-0dev-in-ios-ger%C3%A4ten-ipad-oder-iphone

ChristianMayer:
Fixes #1063 and report at https://knx-user-forum.de/forum/supportforen/cometvisu/1646571-nutzt-jemand-0-12-0dev-in-ios-ger%C3%A4ten-ipad-oder-iphone

### DIFF
--- a/source/class/cv/transforms/Mqtt.js
+++ b/source/class/cv/transforms/Mqtt.js
@@ -59,7 +59,7 @@ qx.Class.define('cv.transforms.Mqtt', {
               ret_pre = '',
               ret_post = '';
             // split on "." but not on "\." to allow the dot to be escaped
-            parameter.split(/(?<!\\)\./).forEach(
+            parameter.match(/(\\\.|[^.])+/g).forEach(
               (e)=>{
                 ret_pre += '{"' + e.replace('\\.', '.') + '":';
                 ret_post += '}';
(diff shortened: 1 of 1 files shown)


source/class/cv/transforms/Mqtt.js
```diff
@@ -59,7 +59,7 @@ qx.Class.define('cv.transforms.Mqtt', {
               ret_pre = '',
               ret_post = '';
             // split on "." but not on "\." to allow the dot to be escaped
-            parameter.split(/(?<!\\)\./).forEach(
+            parameter.match(/(\\\.|[^.])+/g).forEach(
```
peuter, line 62: I do not if it can happen in reality but this code would break if `parameter` is an empty string. So I would suggest to add a null check: `parameter.match(/(\\\.|[^.])+/g)?.forEach(`

source/class/cv/transforms/Mqtt.js
```diff
@@ -73,7 +73,7 @@ qx.Class.define('cv.transforms.Mqtt', {
           let json = JSON.parse(str);
           if( typeof parameter === 'string' ) {
             // split on "." but not on "\." to allow the dot to be escaped
-            parameter.split(/(?<!\\)\./).forEach(
+            parameter.match(/(\\\.|[^.])+/g).forEach(
```
peuter, line 76: same here

peuter (changes requested):

peuter (approved):

PR #1110, opened 2022-01-17
do not use page type in navbars because the content

### LINKED ISSUE #1106: 0.12.0-RC1: navbar broken at demo_flat
grafik

peuter:
of a navbar is always treated like text page content.

In detail: All widgets on 2d pages get the style "position: absolute" which does not make sense for
navbar content.

closes #1106


peuter:
I wonder why this ever worked before. @ChristianMayer Do you have any idea?

ChristianMayer (approved):

Issue #1164, opened 2022-01-23
0.12.0-any: Installation of the development version outdated


ChristianMayer:
The instruction about installation of the development version ( -> https://www.cometvisu.org/CometVisu/de/0.12/manual/install/install-dev.html ) are really outdated and need an update

ChristianMayer:
See also #1004

PR #1165, opened 2022-01-23
Remove wiregate from the documentation as it's too old

### LINKED ISSUE #1157: 0.12.0-RC4: Wiregate is too old
As the Wiregate comes with PHP 5.2.6 it's too old for the requirements of the new manager.
The documentation should be updated accordingly.

ChristianMayer:
Remove wiregate from the documentation as it's too old to be supported with the 0.12 release.

Closes #1157 

#1164 should build on top of this PR, but that's a different topic.
Also the new `source/rest/manager/environment.php` should be used to give the user in the manager a hint about not fulfilled requirements.


source/rest/manager/environment.php
```diff
@@ -0,0 +1,37 @@
...
+$retval = array();
+$retval['phpversion'] = phpversion();
+if (!defined('PHP_VERSION_ID')) {
+  $version = explode('.', PHP_VERSION);
+  define('PHP_VERSION_ID', ($version[0] * 10000 + $version[1] * 100 + $version[2]));
+}
+$retval['PHP_VERSION_ID'] = PHP_VERSION_ID;
+$retval['SERVER_SIGNATURE'] = $_SERVER['SERVER_SIGNATURE'];
+$retval['SERVER_SOFTWARE'] = $_SERVER['SERVER_SOFTWARE'];
+
```
peuter, line 35: Can you please read and add this information here: https://github.com/CometVisu/CometVisu/blob/develop/source/rest/manager/composer.json#L10

We can use that to show the current version requirements in the yet to create error dialog that opens when the PHP version is too old.

doc/manual/de/install/update-other.rst
```diff
@@ -1,44 +1,44 @@
...
+Update der direkt installierten CometVisu
+=========================================
 
 **Download der Installationsdatei**
 
 .. code-block:: bash
 
     cd /var/www
-    wget -O CometVisu0.9.2.zip  https://github.com/CometVisu/CometVisu/releases/download/v0.9.2/CometVisu-0.9.2.zip
+    wget -O CometVisu0.12.0.zip  https://github.com/CometVisu/CometVisu/releases/download/v0.12.0/CometVisu-0.12.0.zip
```
peuter, line 11: Shall we really use real versions here? They need to be kept up to date manually. Something that we forget easily.

peuter (changes requested):

doc/manual/de/install/update-other.rst
ChristianMayer: That's something we also forgot in the past :)
I'd still keep it as an example - and in the hope that we aren't forgetting it too often.

Best would be a script that it automating that - but as we are releasing not so often, a script must also be kept running and an a bit outdated information here doesn't hurt I'd keep it as it is

source/rest/manager/environment.php
```diff
@@ -33,5 +33,9 @@
 $retval['SERVER_SIGNATURE'] = $_SERVER['SERVER_SIGNATURE'];
 $retval['SERVER_SOFTWARE'] = $_SERVER['SERVER_SOFTWARE'];
 
+$composer_file = file_get_contents("composer.json");
+$composer = json_decode($composer_file, true);
+$retval['reqired_php_version'] = $composer['require']['php'];
```
peuter, line 38: typo in array key => `required_php_version`

peuter (changes requested):

PR #1170, opened 2022-01-25
read environment.php from backend and disable manager when PHP versio…

peuter:
…n does not match the constraint

### DIFF
--- a/source/class/cv/Application.js
+++ b/source/class/cv/Application.js
@@ -863,7 +863,7 @@ qx.Class.define('cv.Application',
       xhr.addListenerOnce('success', function (e) {
         const req = e.getTarget();
         const env = req.getResponse();
-        const [major, minor, patch] = env.phpversion.split('.').map(ver => parseInt(ver));
+        const [major, minor] = env.phpversion.split('.').map(ver => parseInt(ver));
         const parts = env.required_php_version.split(' ');
         const disable = parts.some(constraint => {
           const match = /^(>=|<|>|<=|\^)(\d+)\.(\d+)$/.exec(constraint);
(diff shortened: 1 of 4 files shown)


source/translation/de.po
```diff
@@ -1055,3 +1055,11 @@ msgstr "Verwerfen & schließen"
...
 msgid "Save & quit"
 msgstr "Speichern & beenden"
+
+#: cv/Application.js
+msgid "Manager is not available"
+msgstr "Der Manager steht nicht zu Verfügung"
+
+#: cv/Application.js
+msgid "Your system does not provide the required PHP version for the manager. Installed: %1, required: %2"
+msgstr "Die auf Ihrem System installierte PHP version wird vom Manager nicht unterstützt. Installiert: %1, erforderlich: %2"
```
ChristianMayer, line 1065: version -> Version

source/class/cv/Application.js
```diff
@@ -822,6 +853,69 @@ qx.Class.define('cv.Application',
...
+              case '<=':
+                if (major > majorConstraint || (major === majorConstraint && minor > minorConstraint)) {
+                  return true;
+                }
+                break;
+              case '<':
+                if (major >= majorConstraint || (major === majorConstraint && minor >= minorConstraint)) {
+                  return true;
+                }
+                break;
```
ChristianMayer, line 894: (Without explicit testing:) I guess those rules are not correctly implemented. At least `>=` and `>` as well as `<=` and `<` seem to be wrongly exchanged with each other.

As the environment.php is also exporting `PHP_VERSION_ID`: it might be a simpler way to compare the versions.

ChristianMayer (changes requested):

source/class/cv/Application.js
peuter: The comparisons work, please keep in mind that `return true` means the version constraint fails. e.g. `>=7.0` fails when major < 7 OR (major === 7 AND minor < 0)

Using the version id might be a better option, I will check tha.

ChristianMayer (approved):

PR #1174, opened 2022-01-28
Upgrade backend framework to slim4

peuter:
replace the deprecated openapi php-slim generator with php-slim4.

This is not meant to be cherry-picked into the 0.12 release. Needs to be tested a while in real environments and is nor ready for a release.


ChristianMayer:
I'd like to postpone the pull a bit till 0.12.0 is more stable so that it's easier to fix 0.12.0 issues in develop and cherry pick them then

ChristianMayer:
What is the real PHP requirement? 
With my develop container (derived from the CometVisuAbstractBase as well as the official container, but probably not from the most recent one) I get this message when I try to open the manager:
`Die auf Ihrem System installierte PHP Version wird vom Manager nicht unterstützt. Installiert: 8.0.16, erforderlich: ^7.4 || ^8.1`

peuter:
I changed the requirement to by in sync with the one the slim framework has documented `PHP 7.4 or newer`.

peuter:
Are there any open questions left or can we finally merge this?

ChristianMayer:
Thanks for the bump. Currently no open questions, I just lost track of it staying knee deep in MySQL to fix #1281 :)

I'll try this PR ASAP to give it the review.

ChristianMayer (changes requested):
using this PR, opening the manager I get many HTTP 500 results in the network pane. The log of the web server has quite a few lines like:
```

[Sat Jul 23 20:22:18.379547 2022] [php:error] [pid 5187] [client 172.17.0.1:47170] PHP Fatal error:  Declaration of OpenAPIServer\\Container::get(string $id) must be compatible with Psr\\Container\\ContainerInterface::get($id) in /var/www/html/obiwan/rest/manager/src/Container.php on line 19, referer: https://timberwolf76.local/proxy/visugit/obiwan/?config=demo_flat
```

peuter:
Did you run `composer install` in the `source/rest/manager/` ?

ChristianMayer:
No - I expected that `npx qx compile --watch --clean` is sufficient.
But doing so fails:
```
$ composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - slim/slim is locked to version 3.12.3 and an update of this package was not requested.
    - slim/slim 3.12.3 requires ext-simplexml * -> it is missing from your system. Install or enable PHP's simplexml extension.

To enable extensions, verify that they are enabled in your .ini files:
    - /etc/php/8.0/cli/php.ini
    - /etc/php/8.0/cli/conf.d/10-opcache.ini
[...]
You can also run `php --ini` inside terminal to see which files are used by PHP in CLI mode.
```

peuter:
are you sure that your composer.lock file is equal to the one in this PR? There is no lock to slim 3.12.3 in there, at least I can't find one. Maybe deleting the lock file might help.

ChristianMayer:
Running it from my local system (Kubuntu 22.04), not from the container:
```
$ rm composer.lock 
$ composer install
[...]
No lock file found. Updating dependencies instead of installing from lock file. Use composer update over composer install if you do not have a lock file.
Loading composer repositories with package information
[...]
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php >=7.0 <=8.0 but your php version (8.0.8) does not satisfy that requirement.
```
Running it from the container doesn't work as `composer` is not installed - and it's actually nothing that I want to have in the container as it should be as lightweight as possible

peuter:
This PR changes the PHP requirement from `>=7.0 <=8.0` to `>=7.4`. So it seems that you are not operating with the data from this PR.

ChristianMayer:
Installing some PHP packages on my local system `composer install` worked.

But: `npx qx compile --watch --clean` fails with:
```
node:fs:585
  handleErrorFromBinding(ctx);
  ^

Error: EPERM: operation not permitted, open 'compiled/source/ServiceWorker.js'
    at Object.openSync (node:fs:585:3)
    at Object.readFileSync (node:fs:453:35)
    at CvCompileHandler.updateCacheVersion (/home/cm/devel/CometVisu/CometVisu/utils/compile/AbstractCompileHandler.js:76:22)
    at async CvCompileHandler._onMade (/home/cm/devel/CometVisu/CometVisu/utils/compile/cv/CompileHandler.js:84:7) {
  errno: -1,
  syscall: 'open',
  code: 'EPERM',
  path: 'compiled/source/ServiceWorker.js'
}

Node.js v17.3.1
```
And opening the manager gives a few HTTP 500 with the apache log saying:
```
[Sun Jul 24 19:47:57.697548 2022] [php:error] [pid 19435] [client 172.17.0.1:58138] PHP Fatal error:  Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.1.0". You are running 8.0.16. in /var/www/html/obiwan/rest/manager/vendor/composer/platform_check.php on line 24, referer: https://timberwolf76.local/proxy/visugit/obiwan/?config=demo_flat
```

(K)ubunung 22.04 comes with
```
$ php -v
PHP 8.1.2 (cli) (built: Jun 13 2022 13:52:54) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.2, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.2, Copyright (c), by Zend Technologies
```

peuter:
There must be some kind of right problem with the ServiceWorker.js file. Check the owner and rights, maybe delete it by hand an re-compile.

And I guess it's not a good idea to run composer install with a different php version on one system and execute the code with a lower php version on another system. I'm no composer or php expert but if you would do that with npm it most likely wouldn't work either.

ChristianMayer:
It'll be always like that: one system with its own tool versions is creating the files and a completely different system (e.g. the Docker container, or the totally unknown system that runs the final CV) will run the files.

When that's not working it's not a robust solution and will create us a support nightmare.

peuter:
It hasn't been a problem in the past and is it has been this way since we have the new PHP backend. So it must be a problem on your setup. I just did another try (delete composer-lock, vendors subdir and run composer install) and my system has other minimal requirements for the php version than yours: `Your Composer dependencies require a PHP version ">= 8.0.2"`and not `>=8.1.0`. 
I have no idea why that is the case.

All this PR changes is an increase of the minimal required php version from former >=7.4 to >=8.0.2. In Docker this requirement is already fullfilled. Users that install it manually will have to take care about that themselves.

peuter:
Ok composer uses an own cache, and after I deleted it i get the same minimal 8.1.0 requirement. Thats weird and came unexpected. I still don't know where that requirement comes from, the slim framework says >=7.4 is the minimum requirement.

peuter:
Ok this PR also added sentry support for the php backend and that increases the version to >=8.1, I think I will remove the support again as it never worked the way I wanted it. But I will check other dependencies too tomorrow, maybe some of them are also not needed anymore.

peuter:
Please try again it should work with php 8.0 now.

source/rest/manager/README.md
```diff
@@ -1,12 +1,14 @@
...
-* [Slim Framework Documentation](https://www.slimframework.com/docs/)
+* [Slim 4 Documentation](https://www.slimframework.com/docs/v4/)
+
+This server has been generated with [Slim PSR-7](https://github.com/slimphp/Slim-Psr7) implementation.
 
 ## Requirements
 
 * Web server with URL rewriting
-* PHP 7.0 or newer
+* PHP 7.4 or newer
```
ChristianMayer, line 11: Doesn't this need to be bumped to 8.0 now?

source/rest/manager/index.php
```diff
@@ -1,118 +1,73 @@
 <?php
 /**
  * CometVisu Manager backend
+ * PHP version 7.4
```
ChristianMayer, line 4: bump to PHP 8.0?
Is it relevant here or is the central definition of PHP >= 8.0 sufficient?

source/rest/manager/lib/Api/AbstractConfigApi.php
```diff
@@ -1,17 +1,15 @@
 <?php
+
 /**
- * AbstractConfigApi
- *
- * PHP version 7
+ * CometVisu Manager backend
+ * PHP version 7.4
```
ChristianMayer, line 5: bump to PHP 8.0?
Is it relevant here or is the central definition of PHP >= 8.0 sufficient?

source/rest/manager/lib/Api/AbstractDataproviderApi.php
```diff
@@ -1,17 +1,15 @@
 <?php
+
 /**
- * AbstractDataproviderApi
- *
- * PHP version 7
+ * CometVisu Manager backend
+ * PHP version 7.4
```
ChristianMayer, line 5: bump to PHP 8.0?
Is it relevant here or is the central definition of PHP >= 8.0 sufficient?

source/rest/manager/lib/Api/AbstractFsApi.php
```diff
@@ -1,17 +1,15 @@
 <?php
+
 /**
- * AbstractFsApi
- *
- * PHP version 7
+ * CometVisu Manager backend
+ * PHP version 7.4
```
ChristianMayer, line 5: bump to PHP 8.0?
Is it relevant here or is the central definition of PHP >= 8.0 sufficient?

ChristianMayer (changes requested):

source/rest/manager/README.md
peuter: Although this file cah been created by a generator, I changed the version by hand, because this file does not seem to be overriden by new generator call.

source/rest/manager/lib/Api/AbstractConfigApi.php
peuter: This is a comment in generated code, changing it manually would get lost on every new generator execution.

source/rest/manager/lib/Api/AbstractFsApi.php
peuter: This is a comment in generated code, changing it manually would get lost on every new generator execution.

source/rest/manager/lib/Api/AbstractDataproviderApi.php
peuter: This is a comment in generated code, changing it manually would get lost on every new generator execution.

source/rest/manager/index.php
peuter: This is a comment in generated code, changing it manually would get lost on every new generator execution.

ChristianMayer (approved):

PR #1182, opened 2022-01-31
detect mobile value from screen size not from user agent, but

peuter:
still allow the manual override via queryString


peuter:
I am not really sure about if this one should be cherry-picked into 0.12. It is quite a change in the behavior to not longer rely on the user-agent but the current screen size to apply the mobile mode. I think it is an improvement that need to be tested a little but more to get a feeling about the impact it might have.

ChristianMayer (approved):

PR #1183, opened 2022-01-31
Fix icon flicker

### LINKED ISSUE #1026: Icons are too slow / flicker
Having e.g. a button (like a switch) with icon that changes between the two states and the button changes state the icon disappears and then, after a little but noticeable time, the new icon appears.

Expected would be an immediate change from the one to the other icon.

ChristianMayer:
Closes #1026 and also improves code documentation as well a little bit of simplification

### DIFF
--- a/source/ServiceWorker.js
+++ b/source/ServiceWorker.js
@@ -6,7 +6,7 @@
  */
 
 var CACHE = "cv-cache-v2";
-var CACHE_TEST = /.+\.(js|jpg|gif|webp|svg(#.*)?|ttf|woff|eot|css|png|html|json)?$/i;
+var CACHE_TEST = /.+\.(js|jpg|gif|webp|svg(#.*)?|ttf|woff|eot|css|png|html|json)$/i;
 var config = {};
 var updateQueue = [];
 var queueTid = null;
--- a/source/ServiceWorker.js
+++ b/source/ServiceWorker.js
@@ -6,7 +6,7 @@
  */
 
 var CACHE = "cv-cache-v2";
-var CACHE_TEST = /.+\.(js|jpg|gif|webp|svg|ttf|woff|eot|css|png|html|json)$/i;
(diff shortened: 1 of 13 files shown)


peuter (changes requested):
I just did a quick test with loading the demo in testMode `?config=demo&testMode=true` and when I toggle the state of the switch with the name "Switch icon" on the start page it shows a wrong value in off state "nullAus" instead of "Aus". This does not happen in RC5.

peuter (changes requested):
Next problem, when I enable cache and load the demo config (so that cache is used) and toggle the same button it contains this "[object Object]Aus"

peuter:
And when I remove the icon fix from the demo-config (so that it maps to an unkown icon again) I have "[unkown]Aus". But I guess this is the wanted behavior to tell the user that this is an unknown icon?

ChristianMayer:
Yes, "[unkown]Aus" shows the bug in the demo_config that it is using an icon that doesn't exist. In the past it was just ignored, now with cd19b90 this error message is shown to the user - and the demo_config is fixed as well.

peuter (changes requested):
Problem from https://github.com/CometVisu/CometVisu/pull/1183#pullrequestreview-870744290
is still repoducible

source/class/cv/parser/MetaParser.js
```diff
@@ -85,7 +85,9 @@ qx.Class.define('cv.parser.MetaParser', {
 
     parseIcons: function(elem) {
       cv.Config.configSettings.iconsFromConfig.push(this.__parseIconDefinition(elem));
-      qx.core.Init.getApplication().loadIcons();
+      cv.Config.configSettings.iconsFromConfig.forEach(function(icon) {
+        cv.IconHandler.getInstance().insert(icon.name, icon.uri, icon.type, icon.flavour, icon.color, icon.styling, icon.dynamic, icon.source);
+      }, this);
```
peuter, line 90: This seems to be wrong here, it is called after every `icon-definition` entry. Icons should be loaded once after all icon-definitions have been parsed.

peuter (changes requested):

PR #1185, opened 2022-02-01
do not use getElementById, for some reason it does not work when the …

### LINKED ISSUE #1184: 0.12.0-RC5: clock plugin popup doesn't work on iPhone/iPad
See https://forum.timberwolf.io/viewtopic.php?f=95&t=3322&p=37428#p37409

At
```
<clock src="plugins/clock/clock_full.svg" src_popup="plugins/clock/clock_full.svg" title_popup="Testzeit" hide_am_pm="true" hide_am_pm_popup="true" hide_seconds="true" hide_seconds_popup="true" send_on_finish="true" hide_24h="true" hide_24h_popup="true">
	<layout colspan="3" rowspan="2" />
	<address transform="DPT:10.001" mode="readwrite">XX/XX/XX</address>
</clock>
```
the popped up clock doesn't react to touch input on an iPhone or iPad

peuter:
…scg is added to the popup in safari browsers

closes #1184

### DIFF
--- a/source/class/cv/plugins/Clock.js
+++ b/source/class/cv/plugins/Clock.js
@@ -301,15 +301,15 @@ qx.Class.define('cv.plugins.Clock', {
 
           svg.setAttribute('style', 'touch-action: none'); // prevent scroll interference
 
-          let HotSpotHour = svg.getElementById('HotSpotHour');
+          let HotSpotHour = svg.querySelector('#HotSpotHour');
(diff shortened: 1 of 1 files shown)


ChristianMayer (approved):

ChristianMayer:
Probably it's caused by having multiple times the same ID in the page as the SVG is included more than once and it brings those ids with it?

PR #1194, opened 2022-02-05
Expose Timberwolf server info to Sentry

ChristianMayer:


source/class/cv/Application.js
```diff
@@ -992,6 +992,20 @@ qx.Class.define('cv.Application',
             this.info('Manager available for PHP version', env.phpversion);
           }
           this.setManagerChecked(true);
+
+          if (window.Sentry) {
+            Sentry.configureScope(function (scope) {
+              if ('host_release' in env) {
+                scope.setTag('host_release', env.host_release);
```
peuter, line 999: I would prefer if we use the namespace `server.` which is already used e.g. here https://github.com/CometVisu/CometVisu/blob/develop/source/class/cv/TemplateEngine.js#L305

so that would be `server.release`, `server.branch` and `server.id`

peuter (changes requested):

peuter (approved):

PR #1195, opened 2022-02-07
Xml code formatting

peuter:
do not format code when there is a textNode involved, also some minor fixes for the XMLEditor that popped up during testing:

1. Moving an item to a position before a textNode did not work.
2. Fixed an error Message in the console when an SVG-icon is shown in the XML-Editor (the error did not break anything, at least I have not seen a problem)


ChristianMayer:
What problem is fixed by this PR? How can I reproduce it?

peuter:
Edit a config with a label that contains text and an icon and test that the XML-Editor does not change the formatting (indentation) when you save the file.

And with having the textNode as first child and the icon as second, move the icon to the first position: That did not work. The change was not detected and also not saved.

ChristianMayer (approved):

PR #1258, opened 2022-03-22
Fix documentation according to review.

### LINKED ISSUE #1257: 0.12.0-any: documentation: attribute lib_version in description of xml-Header missing
In current documentation releases:

Attribute lib_version is not listed but is required accoding to visu_config.xsd.
Also not shown in example line.
Should be "9" within v0.12.0 according to demo configuration (?).

https://www.cometvisu.org/CometVisu/de/0.12/manual/config/xml-format.html#xml-format-header
https://www.cometvisu.org/CometVisu/en/0.12/manual/config/xml-format.html#xml-format-header

ChristianMayer:
Closes #1257


source/resource/visu_config.xsd
```diff
@@ -413,10 +413,11 @@
         <xsd:enumeration value="cgi-bin" />
         <xsd:enumeration value="oh" />
         <xsd:enumeration value="oh2" />
+        <xsd:enumeration value="openhab2" />
         <!-- New values, should be used now -->
         <xsd:enumeration value="default" />
+        <xsd:enumeration value="knxd" />
```
peuter, line 419: As default == knxd we should also move default to the deprecated section, right?

peuter (changes requested):

peuter:
have a look at the spellcheck, there are spelling errors in the documentation. These are checked now in PRs, because they slipped through when that was not the case. Although the errors are not introduced with this PR we should get that straight now.

peuter (approved):

PR #1286, opened 2022-10-03
Upgrade to qooxdoo v7

peuter:
Upgrade to latest qx version and "es6ify" the source code by using qooxdoos builtin job for that.
This PR also includes a change to use the qx-builtin tangible theme to get rid of some old (and outdated/unmaintained) dependencies. The theme is used in the manager, and because the tangible theme is based on color calculations it is not possible to get the exact same colors as we have used before. but now the contrast is higher so its even better.

You need to run an `npm install` before testing the changes.


peuter:
Actually this PR needs a new qooxdoo release which contains https://github.com/qooxdoo/qooxdoo/pull/10453, which is not available yet.

ChristianMayer:
The es6ify should support staying with single quotes (-> https://prettier.io/docs/en/options.html#quotes ).
When it's possible to stay with the single quotes I'd prefer that (tradition, less changes lines, ...).

But (as stated before) I wouldn't make a religion out of that. So when there are hard facts to switch to double quotes I wouldn't put a veto here.

peuter:
> The es6ify should support staying with single quotes (-> https://prettier.io/docs/en/options.html#quotes ). When it's possible to stay with the single quotes I'd prefer that (tradition, less changes lines, ...).

If it has been possible it would have done it. But there are no configuration options for that job. So you have to live with the pre-defined settings that qooxdoo uses in their sources, and they use double quotes.
 
> But (as stated before) I wouldn't make a religion out of that. So when there are hard facts to switch to double quotes I wouldn't put a veto here.

There are no hard facts just the avoidance of putting too much effort in finding a way to get that job to use a custom config. I think we can live with that change as its just a convention we can get used too.

peuter:
Having a look again at the es6ify documention this is more complete now that when I looked at it the first time. So if the prettier-config option is working I will give it a shot.

peuter:
I changed my mind on waiting for a new qooxdoo release before merging this one. It is just a tiny error somewhere in the manager that will be introduced with this change (until qx fix is released). 

But this pull request is a show-stopper for further development because of merge-conflicts later. So I vote for merging this ASAP.

ChristianMayer:
OK, then I'll priorize it

client/source/class/cv/io/Client.js
```diff
@@ -475,16 +498,30 @@ qx.Class.define('cv.io.Client', {
             delete options.beforeSend;
           }
           if (options.listeners) {
-            Object.getOwnPropertyNames(options.listeners).forEach(function(eventName) {
-              var qxEventName = eventName !== 'error' ? eventName : 'statusError';
-              ajaxRequest.addListener(qxEventName, options.listeners[eventName], context);
+            Object.getOwnPropertyNames(options.listeners).forEach(function (
+              eventName
+            ) {
```
ChristianMayer, line 503: Can we avoid this type of change?

I think
```
            Object.getOwnPropertyNames(options.listeners).forEach(function (eventName) {
``` 
is far more readable than
```
            Object.getOwnPropertyNames(options.listeners).forEach(function (
              eventName
            ) {
```

client/source/class/cv/io/transport/LongPolling.js
```diff
@@ -247,7 +317,16 @@ qx.Class.define('cv.io.transport.LongPolling', {
...
+
+          this.error(
+            'Error! Type: "' +
+              str +
+              '" ExceptionObject: "' +
+              excptObj +
+              '" readyState: ' +
+              readyState
+          );
+
```
ChristianMayer, line 329: Would look better by using the `-Syntax and stay in one line, i.e.
```
  this.error(`Error! Type: "${str}" ExceptionObject ...`);
```

client/source/class/cv/io/transport/Sse.js
```diff
@@ -56,13 +55,17 @@ qx.Class.define('cv.io.transport.Sse', {
...
       this.sessionId = json.s;
       this.version = json.v.split('.', 3);
 
       if (parseInt(this.version[0]) > 0 || parseInt(this.version[1]) > 1) {
-        this.error('ERROR CometVisu Client: too new protocol version (' + json.v + ') used!');
+        this.error(
+          'ERROR CometVisu Client: too new protocol version (' +
+            json.v +
+            ') used!'
+        );
```
ChristianMayer, line 68: Also: single line and backtick string would be preferred

client/source/class/cv/io/transport/Sse.js
```diff
@@ -132,18 +152,22 @@ qx.Class.define('cv.io.transport.Sse', {
...
-      }.bind(this), false);
+    __addRecordedEventListener(topic) {
+      this.debug('subscribing to topic ' + topic);
+      this.eventSource.addEventListener(
+        topic,
+        function (e) {
+          this.dispatchTopicMessage(topic, e);
+        }.bind(this),
+        false
+      );
```
ChristianMayer, line 163: Another example where I think the older, compatcer style was more easy to read and understand

client/source/class/cv/io/Client.js
peuter: All I can do is changing the allowed line width (default 80) to e.g. 120 to avoid some line breaks. Other options to control this kind of formatting do not seem to be possible, at least I have found nothing in the prettier documentation.

Prettier calls iself an "opinionated code formatter" which implies that you don't have a lot of options to configure it for your own needs.

client/source/class/cv/io/transport/LongPolling.js
peuter: I will not change every concatenated string to use backticks, thats not the purpose of this pull request. Please note that it still would add linebreaks if the line exceeds the allowed width, even with backticks. All I can do is to increase the allowed line width, to avoid some of the line breaks.

source/class/cv/ConfigCache.js
```diff
@@ -87,36 +99,60 @@ qx.Class.define('cv.ConfigCache', {
...
+            name
+          ) {
             const mapping = cv.Config.configSettings.mappings[name];
             if (mapping && mapping.formulaSource) {
-              mapping.formula = new Function('x', 'var y;' + mapping.formulaSource + '; return y;'); // jshint ignore:line
+              mapping.formula = new Function(
+                'x',
+                'var y;' + mapping.formulaSource + '; return y;'
+              );
+              // jshint ignore:line
```
ChristianMayer, line 126: Does that work - or does that comment need to be in the line above?

client/source/class/cv/io/transport/Sse.js
peuter: This case is indeed an opinion that prettier and I don't share. And also it is not possible to avoid this change.

source/class/cv/IconConfig.js
```diff
@@ -31,1023 +29,12350 @@ qx.Class.define('cv.IconConfig', {
...
+          '*': { '*': cv.util.IconTools.svgKUF('control_arrow_left') }
+        }
+      },
+
+      control_arrow_leftward: {
+        '*': {
+          white: '*/white',
+          ws: '*/white',
+          antimony: '*/blue',
+
```
ChristianMayer, line 12341: Probably we should treat this as a special case manually keep or upgrade the old file.

source/class/cv/ConfigCache.js
peuter: As we do not use jshint anymore (replaced by eslint) this comment is obsolete anyways.

source/class/cv/parser/pure/widgets/ColorChooser.js
```diff
@@ -39,40 +38,54 @@ qx.Class.define('cv.parser.pure.widgets.ColorChooser', {
...
-      let w_curve      = xml.getAttribute('w_curve');
-      let w_scale      = xml.getAttribute('w_scale');
+      let b_strength = xml.getAttribute('b_strength');
+      let b_curve = xml.getAttribute('b_curve');
+      let b_scale = xml.getAttribute('b_scale');
+      let w_x = xml.getAttribute('w_x');
+      let w_y = xml.getAttribute('w_y');
+      let w_strength = xml.getAttribute('w_strength');
+      let w_curve = xml.getAttribute('w_curve');
+      let w_scale = xml.getAttribute('w_scale');
```
ChristianMayer, line 84: Is there an option to keep the `=` aligned?

source/class/cv/plugins/Gauge.js
```diff
@@ -28,50 +27,51 @@
...
       init: 0,
       apply: '_applyThreshold'
     },
-    thresholdVisible        : { check: 'Boolean', init: false },
-    autoScroll              : { check: 'Boolean', init: false },
-    valuesNumeric           : { check: 'Boolean', init: false }
+
+    thresholdVisible: { check: 'Boolean', init: false },
+    autoScroll: { check: 'Boolean', init: false },
+    valuesNumeric: { check: 'Boolean', init: false }
```
ChristianMayer, line 74: Also here, keeping the `:` alignment would be nice

source/class/cv/parser/pure/widgets/ColorChooser.js
peuter: No

peuter: These are all available options: https://prettier.io/docs/en/options.html

ChristianMayer:
Files are looking fine now. This evening I'll then run the test.

peuter:
qooxdoo has release v7.3, the errors are gone now.

ChristianMayer (approved):

PR #1289, opened 2022-10-11
New list models

### LINKED ISSUE #1288: [component] Add more data models for List component
- [x] Call list from fritzbox (use PHP backend from plugin)
- [x] Data from Rss/-log plugin (use PHP backend from plugin)
- [x] document new features

peuter:
closes #1288

### DIFF
--- a/doc/manual/de/config/structure-tile/components/list.rst
+++ b/doc/manual/de/config/structure-tile/components/list.rst
@@ -261,6 +261,66 @@ Das Modell für die Mitglieder eines GroupItems, hat folgende Struktur:
         "active": true
     }]
 
+**Mitgelieferte Modelle**
+
+Die CometVisu liefert einige vordefinierte Modell mit, mit denen einige Datenquellen von Plugins eingebunden werden
+können. Bisher sind das die Anrufliste einer Fritz!Box über das :doc:`tr064 <../../widgets/plugins/tr064/index>`-Plugin und die Daten aus der
+:doc:`RSSlog <../../widgets/plugins/rsslog/index>` -Datenbank. Hierfür ist es nicht nötig die Plugins in der Config einzubinden, denn
+es werden nur die Daten abgefragt und ansonsten nichts von den Plugins selbst benutzt.
+
+Beispiel zur Anzeige der Anruferliste aus dem tr064-Plugin:
+
+.. widget-example::
+
+    <settings design="tile" selector="cv-tile">
+        <fixtures>
+            <fixture source-file="source/test/fixtures/tr064_proxy.xml" target-path="resource/plugins/tr064/proxy.php"/>
+            <fixture source-file="source/test/fixtures/tr064_soap.json" target-path="resource/plugins/tr064/soap.php"/>
+        </fixtures>
+        <screenshot name="cv-tr064plugin-list"></screenshot>
+    </settings>
+    <cv-tile>
+        <cv-list rowspan="3" colspan="3" refresh="120">
+            <model class="FritzCallList" parameters="device=tr064device,max=10"/>
+            <template>
+              <li>
+                <div style="float: left; font-size: 1.5em; padding-right: 8px">
+                  <i class="knxuf-phone_call_in" style="color: #268DDA; vertical-align: middle;" when="${Type}=1"/>
+                  <i class="knxuf-phone_missed_in" style="color: #E45F3B; vertical-align: middle;" when="${Type}=2"/>
+                  <i class="knxuf-phone_call_out" style="color: #8BBF68; vertical-align: middle;" when="${Type}=3"/>
+                </div>
+                <div style="overflow: hidden; text-overflow: ellipsis; white-space: nowrap;">
+                  ${Name || Caller}
+                  <div style="font-size: 0.9em; color: #777;">${Date}</div>
+                </div>
+              </li>
+            </template>
+        </cv-list>
+    </cv-tile>
+
+Beispiel zur Anzeige der Eintrage aus dem RssLog-Plugin:
+
+.. widget-example::
+
+    <settings design="tile" selector="cv-tile">
+        <fixtures>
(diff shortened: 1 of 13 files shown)


ChristianMayer:
Does this PR need an update due to the switch to Qx7?

peuter:
I did a rebase here now and found some lint errors regarding the last changes in the qx7 branch. so I enabled that the ci lint fails when there are linter errors (had been disabled to weit for qx7, now we can use it again).

source/resource/visu_config_tile.xsd
```diff
@@ -767,6 +767,24 @@
               </xsd:simpleType>
             </xsd:attribute>
             <xsd:attribute name="filter" type="xsd:string"/>
+            <xsd:attribute name="class">
+              <xsd:annotation>
+                <xsd:documentation xml:lang="en">Special builtin list models that are included in the CometVisu.</xsd:documentation>
+                <xsd:documentation xml:lang="de">Spezielles Listenmodell, die die CometVisu mitliefert.</xsd:documentation>
```
ChristianMayer, line 773: Grammatik-Fehler

source/resource/visu_config_tile.xsd
```diff
@@ -767,6 +767,24 @@
...
+                <xsd:restriction base="xsd:string">
+                  <xsd:enumeration value="FritzCallList"/>
+                  <xsd:enumeration value="RssLog"/>
+                </xsd:restriction>
+              </xsd:simpleType>
+            </xsd:attribute>
+            <xsd:attribute name="parameters" type="xsd:string">
+              <xsd:annotation>
+                <xsd:documentation xml:lang="en">Parameter for the class list models.</xsd:documentation>
+                <xsd:documentation xml:lang="de">Parameter for die class Listenmodelle.</xsd:documentation>
```
ChristianMayer, line 785: `for`

doc/manual/de/config/structure-tile/components/list.rst
```diff
@@ -261,6 +261,66 @@ Das Modell für die Mitglieder eines GroupItems, hat folgende Struktur:
         "active": true
     }]
 
+**Mitgelieferte Modelle**
+
+Die CometVisu liefert einige vordefinierte Modell mit, mit denen einige Datenquellen von Plugins eingebunden werden
```
ChristianMayer, line 266: Modell -> Modelle

ChristianMayer: das zweite "einige" besser entfernen

doc/manual/de/config/structure-tile/components/list.rst
```diff
@@ -261,6 +261,66 @@ Das Modell für die Mitglieder eines GroupItems, hat folgende Struktur:
...
+können. Bisher sind das die Anrufliste einer Fritz!Box über das :doc:`tr064 <../../widgets/plugins/tr064/index>`-Plugin und die Daten aus der
```
ChristianMayer, line 267: "Bisher sind das die Anrufliste" klingt komisch

doc/manual/de/config/structure-tile/components/list.rst
```diff
@@ -261,6 +261,66 @@ Das Modell für die Mitglieder eines GroupItems, hat folgende Struktur:
...
+:doc:`RSSlog <../../widgets/plugins/rsslog/index>` -Datenbank. Hierfür ist es nicht nötig die Plugins in der Config einzubinden, denn
+es werden nur die Daten abgefragt und ansonsten nichts von den Plugins selbst benutzt.
```
ChristianMayer, line 269: Was ist mit der Hidden-Config? Die muss doch vermutlich korrekt befüllt sein, denn sonst weiß die Plugin-Datenquelle ja nicht wo sie nachfragen soll

ChristianMayer (changes requested):
Apart from the comments we should discuss a general design decision: what is a plugin and what isn't.

The idea behind a plugin is (was) to have "fat" code that not everyone needs separated, so that only those who need it have to pay the price for it (load time, data and memory consumption). This also includes cases where you need more than the minimum setup - which traditionally was a static web server that has no PHP. (The "ideal" low end platform was a patched Fritz!Box).

This is the reason why the TR064/FritzBox and RSSLog are plugins - they can not live without PHP and also not everyone needs the functionality.
Including the "listmodel"s for those in the main code and not in a plugin is eroding the rule set for plugins. 
When those should stay in the main code then the consequence should be that TR064 and RSSLog are also moved to the main code and their status as a plugin should be stopped.

peuter:
I am aware of the reasons for moving code into a plugin. The tile-structure currently does not support any of the existing plugins and I currently don't have any plans to change that (although that decision might change in the future). If it is useful I would prefer to provide that functionality "built-in". There are other ways to keep avoid bloating up the code base. For example the Tile-Charts only load the library for rendering a chart if you use it, so there is no need for the user to explicitly "activate" a plugin like its done in the pure-structure. And that is what I want to achieve here -> avoiding that extra step for the user. Also I think charts are a basic requirement for modern smart-home UIs so its a builtin feature of the tile structure. Loading data from external sources is another basic requirement (and that is what the list-models are made for). They do not add much code, so there is no need to move them to an extra peace of code that need to be additionally loaded.

So the only remaining valid reason to move the listmodels in an plugin, is that you need PHP support for that. If you use openHABs internal webserver for example these list models will not work because of that. But that can be detected automatically. I am preparing another pull request with changes for openhab specific stuff (also one for openhabs codebase). That one will enable detecting that the CometVisu is delivered by openHAB. When that is merged, the next step would be to inform the user about that when he tries to use the list models (and of course also document that information) and in addition detecting that there is no PHP support (in non openhab use case) by checking the response of the already existing request to environment.php. Although the last step could already be done right now, I would prefer to do it in the order I just described here.

So long story short: The tile-structure does not support plugins (on purpose), so I cannot move it to one. And the check for PHP-Support that is needed for these list-models will be delivered in future pull requests.

ChristianMayer (approved):

PR #1290, opened 2022-11-05
check for openhab environent and php support.

peuter:
This is a cherry-picked feature taken from a future changeset for the development branch. This is the minimum requirement to make the cometvisu work when its served by the openHAB server. The check is needed because some of the backend URLs are different in openHAB.

### DIFF
--- a/source/class/cv/io/rest/Client.js
+++ b/source/class/cv/io/rest/Client.js
@@ -39,10 +39,8 @@ qx.Class.define('cv.io.rest.Client', {
     getBaseUrl: function () {
       if (!this.BASE_URL) {
         let path = '';
-        const engine = cv.TemplateEngine.getInstance();
-        const clientBackend = engine.visu && typeof engine.visu.getBackend === 'function' ? engine.visu.getBackend() : {};
-        if (clientBackend.resources && clientBackend.resources.rest) {
-          path = clientBackend.resources.rest;
+        if (qx.core.Init.getApplication().isServedByOpenhab()) {
+          path = '/rest/cv';
         } else {
           path = qx.util.Uri.parseUri(window.location.href).directory + 'rest/manager/index.php';
         }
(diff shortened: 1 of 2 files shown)


source/class/cv/io/rest/Client.js
```diff
@@ -39,10 +39,8 @@ qx.Class.define('cv.io.rest.Client', {
     getBaseUrl: function () {
       if (!this.BASE_URL) {
         let path = '';
-        const engine = cv.TemplateEngine.getInstance();
-        const clientBackend = engine.visu && typeof engine.visu.getBackend === 'function' ? engine.visu.getBackend() : {};
-        if (clientBackend.resources && clientBackend.resources.rest) {
-          path = clientBackend.resources.rest;
```
ChristianMayer, line 45: Is this code path only relevant for OpenHAB?
Or could it also be used by KNX, e.g. when served by the Docker container?

peuter: This has always been a mix-up of two different things. The URL of the CometVisus REST-API has nothing to do with the backend used for state handling. e.g. you can use openHAB as "state-backend" and still have the normal PHP-REST-API (when you use the docker container). Only if you serve the CometVisu via openHABs own webserver, you use the REST-API that the cometvisu binding provides (same features as the PHP stuff, but natively implemented in java to make it work in openhab). And fpr that you need a slightly different URL. And thats what this is all about. So you only need to know if the CometVisu is served via openHAB.

ChristianMayer:
Opening the demo config with this PR included creates this error message on the console:
grafik

ChristianMayer:
Now I tried a plain `release-0.12` branch without this PR - and it also shows this message.
But I didn't switch back to Qx 6 for that.

Could that be the cause of this message?

peuter:
> Could that be the cause of this message?

You never know, but I don't think so. I see the same request here, but it gets an answer and is working just fine. As the error show this is an server error (Status 500) you will find an error in you server logs that will describe the problem.
But it has nothing to do with this pull request.

ChristianMayer (approved):

PR #1302, opened 2022-12-27
fix building apiviewer

peuter:


peuter:
Blocked by https://github.com/qooxdoo/qxl.apiviewer/issues/49

peuter:
Although this does not fix the apiviewer we should merge it because currently the apiviewer generation runs into a timeout and that breaks the whole documentation generation. And this pull request fixes the timeout problem.
So then we should have a working documentation generation again.

PR #1316, opened 2023-01-16
improve config preview handling by:

peuter:
* disabling the preview when the file does not exist yet and cannot be created (because the config folder is not writable) or it exists but is not writable
* recognize errors during file creation/update and show them, also disable the preview when an error during file creation happened

### DIFF
--- a/source/class/cv/ui/manager/editor/Tree.js
+++ b/source/class/cv/ui/manager/editor/Tree.js
@@ -194,7 +194,15 @@ qx.Class.define('cv.ui.manager.editor.Tree', {
 
     _maintainPreviewVisibility() {
       const handlerOptions = this.getHandlerOptions();
-      this.setShowPreview(qx.bom.Viewport.getWidth() > 800 && (!handlerOptions || !handlerOptions.noPreview));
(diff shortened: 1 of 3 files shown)


source/class/cv/ui/manager/editor/Tree.js
```diff
@@ -2274,19 +2285,24 @@ qx.Class.define('cv.ui.manager.editor.Tree', {
...
-
-              this.__modifiedPreviewElements.removeAll();
-              this.resetPreviewState();
-              previewFile.resetTemporary();
+            err => {
+              if (err) {
+                // disable preview, because the file could not be created
+                this.setShowPreview(false);
+                this.error(err);
+                cv.ui.manager.snackbar.Controller.error(this.tr('Disabling preview because the preview file could not be created.'));
```
ChristianMayer, line 2294: I didn't run the code but by looking at it: isn't the translation for this string missing?

ChristianMayer (changes requested):

source/class/cv/ui/manager/editor/Tree.js
peuter: Yes you are right

ChristianMayer (approved):

PR #1330, opened 2023-02-19
Demo tile improvements

peuter:
also brings a breadcrumb navigation as new feature. includes fixes/improvements for navigation menus on small screens.
And an update for all npm dependencies is included.


utils/ci/deploy.sh
```diff
@@ -121,10 +121,9 @@ fi
 
 if [[ "$GENERATE_DEMO" -eq 1 ]]; then
   echo "generating test mode build"
-  CV_TAG_RUNTIME=demo CV_TESTMODE=resource/demo/media/demo_testmode_data.json qx deploy --clean -t build -f=false --source-maps --save-source-in-map -o out/de/$VERSION_PATH/demo
-  grunt update-demo-config --base-dir=out/de/$VERSION_PATH/demo
+  qx deploy --clean -t build -f=false --source-maps --save-source-in-map -o out/de/$VERSION_PATH/demo
   # Copy demo-mode to default config
-  cp out/de/$VERSION_PATH/demo/resource/demo/visu_config_demo_testmode.xml out/de/$VERSION_PATH/demo/resource/config/visu_config.xml
+  cp out/de/$VERSION_PATH/demo/resource/demo/visu_config_demo-tile.xml out/de/$VERSION_PATH/demo/resource/config/visu_config.xml
```
ChristianMayer, line 126: I don't think that the tile demo should replace the metal@pure demo.
As those are so different we could offer both

peuter: 1. I do not think that an 11 year old design should be the first thing a possible new user sees as a demo, we need something more modern as the „face“ of the cometvisu, 
2. We can support both, but not without some changes in the code because the old configs do not support the demo-data needed to be tied to a single config file. It has to be „compiled in“ when building the demo and then you have only one source of demo data for every config that will be loaded. So to support multiple demo configs and at least one of them is not a tile config some work needs to be done first.

peuter: oh and 3. I would like to get this demo out as soon as possible to use it as a teaser for the new design and hopfully get some people to play with it. Its time that others do some testing und find bugs that I didn‘t. Thats why I did not put some effort in making the old demo work in parallel yet. I can still do that when this is out in the wild.

ChristianMayer: I don't think that the homepage demo is the right place for people to try it and test it. There only finished stuff should be published.

But we could easily ask people in the timberwolf forum to test a nightly. I guess they would be more than happy to do that as quite a few are waiting for it and the last thing we told them was to keep calm till it's ready to test.

peuter: The demo linked from the homepage is the release version, this PR generates the develop-demo which is not linked anywhere. When this is online I will post the link in the timberwolf forum and and the KNXUF, I think that if someone can see what is possible and click through a demo that should lower the bar a bit and also increase the desire to try it out that just a request in a forum with maybe some screenshots.

ChristianMayer: OK, when this is not replacing the demo linked at the homepage than it's fine.

What we will do / show with the next release is a complete different discussion that must be decided when the release is coming but not now.

ChristianMayer (approved):

PR #1337, opened 2023-02-27
Fixes for openweathermap plugin:

peuter:
* Tile-related
  * add margin when used in popup
  * fix config parsing by converting to real XML
  * Remove items in demo config to avoid line breaks on small screens
* General fixes:
  * Remove code duplicate from constructor that is already in the refresh-mixin and caused double calls, handle case when no refresh is set by calling the refresh method once


ChristianMayer (changes requested):
This PR brakes my simple OWM test config:
```
<?xml version="1.0" encoding="UTF-8"?>
<pages xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" lib_version="9" design="pure" xsi:noNamespaceSchemaLocation="../../visu_config.xsd">
  <meta>
    <plugins>
      <plugin name="openweathermap" />
    </plugins>
    <mappings>
    </mappings>
    <stylings>
    </stylings>
    <statusbar>
    </statusbar>
  </meta>
  <page name="Start">
<openweathermap q="Munich" appid="..." />
  </page>
</pages>
```

peuter:
That should be fixed now.

ChristianMayer (approved):

PR #1342, opened 2023-03-24
Tile fixes and features

peuter:
includes some minor fixes and improvements and as new feature the possibility to mark a widget as outdated (last updated value too old)


client/source/class/cv/io/Client.js
```diff
@@ -192,6 +192,11 @@ qx.Class.define('cv.io.Client', {
       check: 'String',
       nullable: true,
       event: 'changedServer'
+    },
+
+    name: {
+      check: 'String',
+      nullable: true
```
ChristianMayer, line 199: Is this required here, as it is already included in `IClient.js`?

ChristianMayer (approved):

PR #1345, opened 2023-04-08
add multi-backend support to pure structure and recording

peuter:
also fix replaying xhr responses that are delayed

### DIFF
--- a/client/source/class/cv/io/Client.js
+++ b/client/source/class/cv/io/Client.js
@@ -59,6 +59,15 @@ qx.Class.define('cv.io.Client', {
       backendName = cv.io.Client.backendNameAliases[backendName];
     }
     this.backendName = backendName;
+    switch (this.backendName) {
+      case 'default':
+        this._type = 'knxd';
+        break;
+
+      case 'openhab':
+        this._type = 'openhab';
+        break;
+    }
 
     if (backendName && backendName !== 'default') {
       if (typeof backendName === 'object') {
@@ -208,6 +217,7 @@ qx.Class.define('cv.io.Client', {
   members: {
     backend: null,
     backendName: null,
+    _type: null,
     backendLoginUrl: null,
     addresses: null, // the subscribed addresses
     initialAddresses: null, // the addresses which should be loaded before the subscribed addresses
@@ -222,7 +232,7 @@ qx.Class.define('cv.io.Client', {
     __lastError: null,
 
     getType() {
-      return this.backendName;
+      return this._type;
     },
 
     // property apply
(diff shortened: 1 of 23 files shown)


ChristianMayer:
The demo config worked fine, but demo_flat creates lots of error messages when e.g. playing with the color chooser:
grafik

peuter:
Should work now

ChristianMayer:
Error messages have gone, but changing any actor in demo_flat doesn't show a reaction. Neither on the network pane (no "r" are sent) nor on any other actors that use the same address.

Also in the console I get this message that shouldn't be possible as no communication should go to `/cgi-bin/` directly:
grafik

peuter:
The read request should work now, the GET-Request to cgi-bin/ is not reproducable for me.

ChristianMayer:
This works, but the 403 and the direct call to `/cgi-bin/` is still here. 
Going into the debugger I see a strange mixup up knxd and openhab:
grafik

peuter:
Do you have a openhab backend url configured somewhere? What are the cometvisu specific headers that are send with the config response?

ChristianMayer:
Ok, that leads to the right direction:
grafik
But I haven't configured OpenHAB in the Docker container:
grafik

ChristianMayer:
Looking at https://github.com/CometVisu/Docker/blob/master/CometVisuBase/cometvisu-entrypoint#L40-L64 I think this is (currently) an expected behavior

peuter:
Yes I just checked that too, I think we should delete this line https://github.com/CometVisu/Docker/blob/master/CometVisuBase/cometvisu-entrypoint#L40

I will also change this pull request to handle the backend names as comma separated list correctly, I wasn't aware that this is allowed. Then I will only create clients for the backends that are in that list. That should fix your problem too.

ChristianMayer:
Is is safe (backward compatible) to delete this line 40?

peuter:
> Is is safe (backward compatible) to delete this line 40?

I don't remember if we ever had a release where you had to use the CGI-URL variable to configure the path to the openhab backend. But its clearly wrong if someone did that, so I would take the risk to find someone who has this kind of configuration and make him fix it by removing this behavior.

peuter:
btw as CGI_URL_PATH is deprecated you should remove it from your configuration.

ChristianMayer:
> btw as CGI_URL_PATH is deprecated you should remove it from your configuration.

I know, it's in there due to laziness and testing what's happening to a user who didn't remove it :)

ChristianMayer (approved):

PR #1358, opened 2023-08-10
make environment.php run on php 5.2 and

peuter:
an older apache server that handles the redirects differently

refs https://knx-user-forum.de/forum/supportforen/cometvisu/1884208-hilfe-bei-cometvisupdate-auf-dem-wire-gate-f%C3%BCr-linux-behinderten-gesucht

should be tested in a real wiregate if possible, because the actual error described in the post could not be reproduced in the docker container with php-5.2. In there the script just returned an HTTP 500 error which is already handled fine by the cometvisu client. The error in the post seems to happen when the scripts returns HTTP status 200 but has no or wrong content.

With this fix the scripts returns correct data which leads to a hint that the manager is not supported by this php version when the user tries to start it.


peuter:
Closing this, because the only change left in this PR is the change in the development.md. I will copy that change to the other PR.

PR #1359, opened 2023-08-12
Tile improvements

peuter:
cleanup dependencies:
* removed the ones we do not use anymore
* removed grunt-banner because it it not maintained anymore (last release 8 years ago), i have not found a replacement but Intellij can be configured to add a banner to each newly created file.

Some minor Tile-Design improvements

Add more chart data sources: openHAB, influxdb and RRD to be feature equal with pure Diagram-Plugin regarding the sources. Included documentation should provide enough information to be able to test it. So I will not provide anything here and use is as a test if the documentation is missing something.


doc/manual/de/config/structure-tile/components/chart.rst
```diff
@@ -34,14 +34,94 @@ einen Zeitpunkt im Chart wird als Tooltip angezeigt, wenn man mit dem Mauszeiger
...
+``openhab://<Item-Name>`` wobei <Item-Name> ein beliebiger Item-Name in openHAB sein kann, für den historische Daten aufgezeichnet werden.
+Es ist zusätzlich möglich einen spezielle Persistence-Service anzugeben, falls man mehrere benutzt:
+``openhab://<service-id>@<Item-Name>``, gibt man diesen nicht an wird der in openHAB konfigurierte default-Persistence Service benutzt.
+
+**Konfiguration einer InfluxDB-Quelle**
+
+``flux://<organization>@<bucket>/<measurement>/<field>`` Optional können Angaben für die
+`aggregationWindow-Funktion <https://docs.influxdata.com/flux/v0.x/stdlib/universe/aggregatewindow/>`_ als
+Query-Parameter angehängt werden. Diese benötigen den Präfix ``ag-``. Beispiel:
+``flux://cometvisu@cv-bucket/Temperature/value?ag-every=1d&amp;ag-fn=mean`` würde in folgende Flux-Query an die
```
ChristianMayer, line 54: Double-Check: is `&amp` correct in this case - or should it just be `&`?

ChristianMayer (changes requested):

doc/manual/de/config/structure-tile/components/chart.rst
peuter: Just `&` is no valid XML (even in attribute values), it must be `&amp`

ChristianMayer (approved):

PR #1373, opened 2023-10-15
Fix slider to respect steps and fix initialization after start

ChristianMayer:
- Fix slider to respect steps (cf. https://knx-user-forum.de/forum/supportforen/cometvisu/1900423-slider-oh-empf%C3%A4ngt-kontinuierliche-werte-trotz-step?p=1900598#post1900598 )
- Fix initialization where dimensions could have been read before the design was applied (this could cause bad slider handler positions right after start) 
- Fix little typo with mobile screen size

### DIFF
--- a/source/class/cv/TemplateEngine.js
+++ b/source/class/cv/TemplateEngine.js
@@ -338,9 +338,6 @@ qx.Class.define('cv.TemplateEngine', {
           }
         });
         loader.addListenerOnce('stylesLoaded', this.generateManifest, this);
-        loader.addListenerOnce('stylesAndScriptsLoaded', () => {
-          cv.ui.structure.pure.layout.ResizeHandler.invalidateScreensize();
-        }, this);
       }
       // load structure-part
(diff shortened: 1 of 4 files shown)


source/class/cv/TemplateEngine.js
```diff
@@ -338,6 +338,9 @@ qx.Class.define('cv.TemplateEngine', {
           }
         });
         loader.addListenerOnce('stylesLoaded', this.generateManifest, this);
+        loader.addListenerOnce('stylesAndScriptsLoaded', () => {
+          cv.ui.structure.pure.layout.ResizeHandler.invalidateScreensize();
+        }, this);
```
peuter, line 343: This is only needed for pure-structure and therefore should be done in its controller `cv.ui.structure.pure.Controller`. If you use tile-structure this code would produce an error because the ResizeHandler is not even loaded.

peuter (changes requested):

peuter (approved):

PR #1384, opened 2023-12-23
Convert diagram plugin to timeseries

joltcoke:


source/class/cv/plugins/diagram/AbstractDiagram.js
```diff
@@ -271,146 +271,87 @@ qx.Class.define('cv.plugins.diagram.AbstractDiagram', {
...
-    lookupTsCache(ts, start, end, res, forceNowDatapoint, refresh, force, callback, callbackParameter) {
-      const client = cv.io.BackendConnections.getClient();
-      let key;
-      let url;
-      const chartsResource = client.getResourcePath('charts', {
-        src: ts.src,
-        start: start,
-        end: end
-      });
+    lookupTsCache(ts, start, end, series, period, res, forceNowDatapoint, refresh, force, callback, callbackParameter) {
```
peuter, line 283: This whole change breaks several things:

1. All the cv.io.timeseries.* Classes are written for tile configs which configure the diagram queries differently that this plugin. InfluxDB queries will not work
2. The is a client.getResourcePath('charts',...) in the old code that changes the request URL for rrd datasources when openHAB is the datasource to queries its REST interface. This is missing now and getting chart data from openHAB does not work.

peuter (changes requested):
I also tested charts in tile structure and everything seems to still work there.

source/class/cv/plugins/diagram/AbstractDiagram.js
joltcoke: 1. Yes, this is a known problem. Still trying to figure out how to build the query based on the old XML config
2. Need to look into this. What is the specific setup you're referring to? I'm using the code with OpenHAB and it works just fine.

peuter: 2. I tested before you commited `3b21e89e4c300fcdab581b4830e404c777cec550` now it works.

peuter:
Closed due to inactivity, Feel free to re-open it if you want to continue working in this.

PR #1397, opened 2024-02-19
Several minor fixes + new web-widget

peuter:


doc/manual/de/config/structure-tile/widgets/web.rst
```diff
@@ -0,0 +1,48 @@
...
+------------
+
+Das Web-Widget dient zu Anzeige einer externen Webseite. Optional kann diese Seite in regelmäßigen Abständen
+automatisiert neu geladen werden.
+
+.. code-block:: xml
+
+    <cv-web src="http://www.google.de" refresh="120" style="width: 100%; min-height: 600px;" />
+
+Über das optionale ``refresh``-Attribut festgelegt werden in welchem Interval in Sekunden die Seite neu geladen wird.
```
ChristianMayer, line 19: Satz klingt komisch

doc/manual/de/config/structure-tile/widgets/web.rst
```diff
@@ -0,0 +1,48 @@
...
+Größenangaben können über das style-Attribut per CSS festgelegt werden.
+
+Sofern es dabei Probleme gibt kann das Laden der Seite über einen Proxy im Backend benutzt werden indem
```
ChristianMayer, line 22: `indem` -> `in dem`

doc/manual/de/config/structure-tile/widgets/web.rst
```diff
@@ -0,0 +1,48 @@
...
+man das ``proxy``-Attribut auf ``true`` setzt. Hierfür muss jedoch die URL in die "proxy.whitelist" Sektion der
+:ref:`Versteckten-Konfiguration <hidden-config>` hinzugefügt werden.
+Dafür muss ein neuer Eintrag in der Sektion angelegt werden, die Schlüssel kann beliebig gewählt werden
```
ChristianMayer, line 25: der Schlüssel

source/class/cv/io/timeseries/FluxSource.js
```diff
@@ -167,6 +167,9 @@ qx.Class.define('cv.io.timeseries.FluxSource', {
         lineEntries = line.split(',');
         if (lineEntries[valueIndex]) {
           value = parseFloat(lineEntries[valueIndex]);
+          if (isNaN(value)) {
+            value = 0;
+          }
```
ChristianMayer, line 172: Is `NaN` really equal to `0` - or should this value be skipped as (e.g.) it doesn't exist?

ChristianMayer (changes requested):

source/class/cv/io/timeseries/FluxSource.js
peuter: Unfortunately I had this fix in my local system for a long time, so I do not remember how to reproduce the problem, all I remember that this causes an error. So I do not have an answer if setting to 0 or leaving it out is the better option. Usually chart do not handle gaps well, or you don't event see that there is a gap/problem because the line from the former to the next value just goes over it. So for now I would go with the 0 because you can see the spikes in the chart (although its technically the wrong value)

doc/manual/de/config/structure-tile/widgets/web.rst
```diff
@@ -0,0 +1,48 @@
...
+    "proxy.whitelist": {
+        "server": "https://www.google.de"
+    }
+
+Beispiel mit regulärem Ausdruck (beginnt und endet mit "/"), welcher alle URLs erlaubt, die "webcam" enthalten:
+
+.. code-block:: json
+
+    "proxy.whitelist": {
+        "server": "/^.+google\/.*$/"
```
mivola, line 41: should it be `webcam` instead of `google` in the RegEx?

ChristianMayer (approved):

PR #1399, opened 2024-02-23
Improve cv-button trigger mode, cv-address subscription that are added later and some typos

peuter:


client/source/class/cv/io/Client.js
```diff
@@ -90,6 +90,8 @@ qx.Class.define('cv.io.Client', {
     this.pass = '';
     this.device = '';
     this.headers = {};
+
+    this.delayedRestart = qx.util.Function.debounce(this.restart.bind(this), 50);
```
ChristianMayer, line 94: Especially the client should be as much native JS and as little Qx as possible IMHO.
So is the use of this Qx helper function really necessary or could it be simply replaced by a `setTimeout` or similar?

source/class/cv/io/mqtt/Client.js
```diff
@@ -129,9 +130,17 @@ qx.Class.define('cv.io.mqtt.Client', {
      *
      */
     subscribe(addresses, filters) {
+      this.addresses = addresses ? addresses : [];
```
ChristianMayer, line 133: Wouldn't be the more modern `addresses ?? []` be better?

source/class/cv/ui/structure/tile/components/Select.js
```diff
@@ -97,25 +127,47 @@ qx.Class.define('cv.ui.structure.tile.components.Select', {
     },
 
     _updateValue(mappedValue, value) {
-      if (this.__options.has(mappedValue)) {
+      const key = typeof mappedValue !== 'undefined' ? '' + mappedValue : '';
```
ChristianMayer, line 130: Would this work also and be better?
```
      const key = '' + (mappedValue ?? '');
```

ChristianMayer:
The points I found wouldn't block a pull => @peuter please change them or tell me that you want to stay with them

client/source/class/cv/io/Client.js
peuter: debounce !== setTimeout, I do not see any advantage of inventing the wheel twice when the framework you are using already provides a solution and you gain not a single percent less dependency by using it.
Everything is build and based on Qx so you already have 100% dependency, you can't get more dependency than that. Avoiding the usage of qx-stuff would only make sense if we had plans to remove qx at all. I do not have any plans for that, because that would be a enormous task with no real benefit.

source/class/cv/io/mqtt/Client.js
peuter: Better? No, its just syntax and I'm old school ;-)

source/class/cv/ui/structure/tile/components/Select.js
peuter: IMHO thats not the same e.g. 0 would be converted to an emtpy string, `false` also. The existing line should convert everything to string but avoid `null` and `undefined` as strings, thats what the type check is for.

peuter:
I explained for every point why I want to keep it as it is, so please merge if you have finished your review

ChristianMayer (approved):

PR #1420, opened 2024-12-17
Fix startup and code style

ChristianMayer:
Gracefully handle a too slow start
Don't use the deprecated `unload` page event any more
Update code style and add more annotations

### DIFF
--- a/source/class/cv/TemplateEngine.js
+++ b/source/class/cv/TemplateEngine.js
@@ -153,6 +153,8 @@ qx.Class.define('cv.TemplateEngine', {
      * @param parts {String[]|String} parts to load
      */
     loadParts(parts) {
+      let continueWhenSuccessful = true;
+
       if (!Array.isArray(parts)) {
         parts = [parts];
       }
@@ -177,8 +179,10 @@ qx.Class.define('cv.TemplateEngine', {
                 if (!cv.Config.loadedStructure) {
                   cv.Config.loadedStructure = part.substring(10);
                 }
-                this.debug('successfully loaded all structures');
-                qx.core.Init.getApplication().setStructureLoaded(true);
+                this.debug('successfully loaded all structures. Continue:', continueWhenSuccessful);
+                if (continueWhenSuccessful) {
+                  qx.core.Init.getApplication().setStructureLoaded(true);
+                }
               }
               this.__partQueue.remove(part);
               waitingFor.remove(part);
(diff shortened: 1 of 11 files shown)


source/class/cv/TemplateEngine.js
```diff
@@ -341,7 +348,16 @@ qx.Class.define('cv.TemplateEngine', {
...
-      await this.loadParts([cv.Config.getStructure()]);
+      try {
+        await this.loadParts([cv.Config.getStructure()]);
+      } catch (e) {
+        // Note: the timeout can be changed by the not published URL parameter
+        // timeoutStructureLoad for debugging reasons. Usually, the server
+        // must be quick enough so that the client doesn't run into any issues
+        // here.
+        this.__showFatalError(`${qx.locale.Manager.tr('loadParts "Structure" failed')}: ${e}`);
+        throw new Error('loadParts "Structure" failed');
```
peuter, line 359: this seems redundant. in a non debug build the global error handler might also capture the thrown exception. I think it also created a error notification for that an we have it twice.

peuter (changes requested):

source/class/cv/TemplateEngine.js
ChristianMayer: I wanted to stop the program flow here as it doesn't make sense to continue.
And it shows the source file line on the console.

ChristianMayer:
I'm just seeing that the change to `pagehide` created this error message:
grafik

Is it possible that `qx.event.Registration` can't handle the `pagehide`?

peuter:
Looks as if those events are not handled yet: https://github.com/qooxdoo/qooxdoo/blob/97ef488c2dea6a21edcbefb13f2c7505f4b34c56/source/class/qx/event/handler/Window.js#L66-L75

But I don't see a benefit for using the qx event registration here. We can just use the pure native way to listen to that event. And btw. documentation does not sound that this event is much better than "unload". Isn't it just triggered when the user used the browser history to navigate to another page?
I'm not even sure if we need this at all. When the page/cometvisu gets unloaded all clients will stop anyways.

ChristianMayer:
`unload` is deprecated and shouldn't be used - so it's giving a warning in the log.
Looking at https://developer.chrome.com/docs/web-platform/page-lifecycle-api?hl=de#the-unload-event the best replacement was the `pagehide`.

Actually the page lifecycle API is great as it allows us to better handle the backend connection in all different conditions where the page is hidden and shown again. But that's something for a different PR.

peuter:
Please resolve the merge conflicts, otherwise I can't merge this.

ChristianMayer:
It doesn't have a merge conflict (any more).

What's open are failed unit tests - but I actually don't know why they are failing. It looks unrelated to this PR

peuter (approved):

PR #1421, opened 2024-12-17
Fix multi backend for pure

ChristianMayer:
- Correctly handle when multiple backends are configured in the config file `backend` property
- Fix XSD to have the rule that checks for the proper multi backend config here
- show message about unused backend only for debug to prevent false alarms (-> https://knx-user-forum.de/forum/supportforen/cometvisu/1999022-messwerte-oder-stati-in-cometvisu-anzeigen?p=2000369#post2000369)
- simplify `sendToBackend` (change should be neutral)

Note: this PR is based on the changes of #1420 - so that PR should be merged first.

### DIFF
--- a/source/class/cv/data/Model.js
+++ b/source/class/cv/data/Model.js
@@ -162,7 +162,7 @@ qx.Class.define('cv.data.Model', {
           }
         }, this);
       } else {
-        this.warn('no addresses registered for backend "' + backendName + '", skipping update');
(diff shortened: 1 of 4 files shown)


peuter:
Can you please rebase this? I still see all changes from #1420 in the diff. That makes it hard to see the real changes. I assume a rebase to the updated develop branch should clear this up.

peuter (approved):

PR #1424, opened 2024-12-26
Handle a not found plugin better

ChristianMayer:
This PR is a first step but requires a second one for even better handling like really telling the user which file wasn't found

### DIFF
--- a/source/class/cv/util/ScriptLoader.js
+++ b/source/class/cv/util/ScriptLoader.js
@@ -256,7 +256,9 @@ qx.Class.define('cv.util.ScriptLoader', {
         loader.removeListener('loaded', this._onLoaded, this);
         loader.removeListener('failed', this._onFailed, this);
       });
-      loader.start();
+      loader.start(); // NOTE: Actually we'd need a `.catch()` here when the
+      // loading of a plugin is failing. But then no error message at all is
+      // shown. Why?!? FIXME!
       return loader;
     },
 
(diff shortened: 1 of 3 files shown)


source/class/cv/util/ScriptLoader.js
```diff
@@ -256,7 +256,9 @@ qx.Class.define('cv.util.ScriptLoader', {
         loader.removeListener('loaded', this._onLoaded, this);
         loader.removeListener('failed', this._onFailed, this);
       });
-      loader.start();
+      loader.start(); // NOTE: Actually we'd need a `.catch()` here when the
+      // loading of a plugin is failing. But then no error message at all is
+      // shown. Why?!? FIXME!
       return loader;
```
peuter, line 265: The error is sent as notification for topic `cv.loading.error` but no one is listening on this topic.

When you add the `catch(e => {})` here and:

```js
        'cv.loading.error': {
          type: 'error',
          icon: 'message_attention'
        },
```

here https://github.com/CometVisu/CometVisu/blob/develop/source/class/cv/ui/PopupHandler.js#L58 the error is shown as popup.

I guest we should add that to this PR because currently it does not fix anything.

peuter (changes requested):

peuter (approved):

peuter:
Would be better to fix the lint issues instead of overruling them. Lint will fail from now on for every run.

ChristianMayer:
Ah, I didn't notice that the lint was still open and just assumed that you approved and forgot to click merge :)

PR #1426, opened 2024-12-27
Fixes for documentation and new tile features

peuter:
re-generated all en screenshots because language is now part of the calculated hash. So the hash has changed, but the actual screenshot did not.
Also some includes some fixes for documentation.

New features:
* Link widget
* Dynamic content widget
* openHAB scene trigger support
* gradient for charts


doc/manual/en/config/_static/oh_datetime.png
ChristianMayer, line 1: Is it possible to change the locale to EN for generating this image?
And even then `Uhr` should be `o'clock` - and even then it's probably better to just leave it away or replace it with a precedenting `time: `

ChristianMayer (approved):
The issue in `doc/manual/en/config/_static/oh_datetime.png` should be fixed, but that can be done independently from this PR

PR #1443, opened 2025-11-23
feat: use new docker base image

peuter:
also includes improvements for the distrobox that also uses that base image


doc/manual/de/tutorial/rpi_cometvisu.rst
```diff
@@ -189,37 +189,35 @@ Zum ersten mal wird der CometVisu-Server nun gestartet mit: ::
...
+    Durch ``80:80`` wird der Port 80 im Container auf den Port 80 des
+    Hosts, also des Raspberry Pi, gemappt. Erst hierdurch kann auf den
+    Web-Server im Container von außen zugegriffen werden.
+
+``-v``
+    Mit ``CometVisuConfig:/var/www/html/resource/config`` wird das vorhin
+    angelegte Volume ``CometVisuConfig`` auf den Pfad
+    ``/var/www/html/resource/config`` im Container gemappt.
+
+``cometvisu/cometvisu:latest-arm``
```
ChristianMayer: When we are changing this here we should remove the `-arm` as we have a unified container now

doc/manual/de/install/docker.rst
```diff
@@ -40,25 +40,34 @@ Umgebungsvariablen (environment)
...
-|                     |                              |openHAB Backend.                                                 |
-+---------------------+------------------------------+-----------------------------------------------------------------+
+.. list-table::
+   :header-rows: 1
+
+   * - Parameter
+     - Default
+     - Beschreibung
+   * - KNX_INTERFACE
+     -
```
ChristianMayer, line 50: ``iptn:172.17.0.1:3700`` is missing.

ChristianMayer (changes requested):

doc/manual/de/install/docker.rst
peuter: I removed the default value on purpose. The knxd in the container is only started when that value is set, it should not start with a random interface value, that most likely does not work because the user needs to set it for his setup to another value.

ChristianMayer (approved):

ChristianMayer:
Merge it on probation to be better able to check the new knxd for compatability

PR #1444, opened 2025-11-30
More tests for missing tile code + start replacing deprecated protractor

peuter:
This adds AI generated tests for tile code. Also some manual adjustments / fixes that improve the robustness of tests in general.

Also this introduces a new way of screenshot generation (also started by AI and then manually improved). The currently used protractor is deprecated for years now. It is replaced with playwright. Also found some problems around duplicated and improved the new generator to detect them and abort.

Also a small color scheme change for tile that inverts foreground / background colors. The background is now darker then the foreground (before it was the other way around). This is a test and might also be reverted, made optional or improved


peuter:
Ping

ChristianMayer (approved):

PR #1454, opened 2026-06-28
Enhance release

ChristianMayer:
- allow alpha and beta builds
- don't tag a Docker with `latest`  when it is not a final release


utils/ci/docker_push.sh
```diff
@@ -44,7 +44,12 @@ if [[ $IS_TAG == 1 ]]; then
       MASTER_TAG=testing
       TESTING=1
     elif [[ $IN_MASTER -ne 0 ]]; then
-      MASTER_TAG=latest
+      if [[ $TAG == *-* ]]; then
```
peuter: We already have an IS_RELEASE variable, please use that. latest tag only when $IS_RELEASE -eq 1

peuter (approved):

PR #1455, opened 2026-06-28
fix: improve handling of non existing custom_visu_config.xsd

peuter:
default file has been removed from user-specific config folder to make sure that its always shipped with the cometvisu code. For our own validation in the manager there is a fallback implemented to load the default file when there is none in the config-folder.

Also there has been a check added to the environment.php that checks if the custom-schema file exists in the config folder when there are configs in there that use the tile-structure (otherwise that file is not needed). Use is asked then to automatically create it or he can ignore the hint and won't be asked again. If he allows it it will be copied from the shipped default file (copy function needed to be added to the filsystem-api, also a mount-point for that specific file because its outside the accessigle folders).

The third change automatically copied the default file whenever the user creates a new tile-config-file.


source/rest/manager/src/Api/FsApi.php
```diff
@@ -273,6 +273,60 @@ public function move(
         }
     }
 
+    public function copy(
```
ChristianMayer, line 276: The paths should be restricted with a white list to prevent any bad security issue where a system file might be copied and thus made visible.

Generally only the user config directory should be available for read and write operations. Selected additional source could be possible.

ChristianMayer (changes requested):

source/rest/manager/src/Api/FsApi.php
peuter: They already are, you cannot do anything with files outside the config + demo-folders. Thats why that special mount for the custom_visu_config.xsd was needed, without it you could not access the file.

ChristianMayer (approved):

PR #1485, opened 2026-09-04
Show already received values while other addresses are still missing

joltcoke:
## Problem

A widget whose `format` refers to several addresses only showed `-` until the last address had
been received, and logged a warning on **every** update until then:

```
cv.ui.structure.pure.Info[908-0]: TypeError: [sprintf] expecting number but found undefined,
["%1$.1f °C   ==>   %2$.1f °C",19]
```

Example configuration:

```xml
<info format="%1$.1f °C   ==&gt;   %2$.1f °C">
  <address format-pos="1">Outside</address>
  <address format-pos="2">Exhaust</address>
</info>
```

## Cause

`applyFormat()` fills `formatValueCache` per address and calls sprintf immediately. Since the
values of the addresses arrive one by one, the arguments of the other positions are still
`undefined`, which sprintf rejects for its numeric conversions. `cv.util.String.sprintf` catches
that, logs the warning and returns `-` for the whole widget.

## Fix

Instead of the value, the **placeholder** of a missing position is replaced by `-` before the
format is handed to sprintf. The values that already arrived are therefore visible right away:

| state | before | after |
| --- | --- | --- |
| only position 1 received | `-` | `19.0 °C   ==>   - °C` |
| both received | `19.0 °C   ==>   21.0 °C` | unchanged |

Substituting the value itself is no option: the numeric conversions of sprintf only accept real
numbers, `'-'` and `null` throw, `NaN` renders as `NaN` and `0` would look like a real reading.

Formats without explicit positions (`%.1f`) are untouched - their argument is consumed
sequentially and cannot be attributed to an address.

## sprintf-js compatibility

The expression that finds the placeholders is a copy of the one inside sprintf-js and could drift
apart when that library changes. A spec therefore compares it against `sprintf.parse()` of the
installed version, builds its corpus from the conversion characters that version actually accepts
and reads the parse tree in both shapes (1.0.x hands out the raw match array, 1.1.x an object).
A syntax change - 1.1.x added `t`, `T` and `v` - fails the tests instead of silently mis-parsing.

## Tests

`grunt karma:ci`: 1741 passing. The new specs cover a partially received format, a fully received
one, a later update of a single address, string placeholders, a literal `%%`, single value
formats and the equivalence with sprintf-js. Removing a character from the expression makes the
equivalence spec fail, so it actually guards the copy.

### DIFF
--- a/source/class/cv/ui/common/BasicUpdate.js
+++ b/source/class/cv/ui/common/BasicUpdate.js
@@ -82,7 +82,17 @@ qx.Mixin.define('cv.ui.common.BasicUpdate', {
      * against the parser of the installed sprintf-js, so such a change fails the tests. When that
      * happens, align this expression with the "placeholder" regex of the new version.
      */
-    PLACEHOLDER_REGEX: /%%|%(?:([1-9]\d*)\$|\(([^)]+)\))?(\+)?(0|'[^$])?(-)?(\d+)?(?:\.(\d+))?([b-gijosuxX])/g,
+    PLACEHOLDER_REGEX: /%%|%(?:([1-9]\d*)\$|\(([^)]+)\))?(\+)?(0|'[^$])?(-)?(\d+)?(?:\.(\d+))?([b-gijosuxX])/,
+
+    /**
+     * A fresh global regex for the placeholder syntax. A global regex keeps its own lastIndex, so a
+     * shared instance would make callers depend on each other's search progress.
+     *
+     * @return {RegExp} a copy of {@link cv.ui.common.BasicUpdate#PLACEHOLDER_REGEX} with the g flag
+     */
+    placeholderRegex() {
+      return new RegExp(cv.ui.common.BasicUpdate.PLACEHOLDER_REGEX.source, 'g');
+    },
 
     /**
      * Replace those placeholders of a format string whose value is still missing by a literal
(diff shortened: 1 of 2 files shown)


source/class/cv/ui/common/BasicUpdate.js
```diff
@@ -63,6 +63,64 @@ qx.Mixin.define('cv.ui.common.BasicUpdate', {
...
+     * The placeholder syntax of sprintf-js, taken from its own source (1.0.3) so that every case it
+     * accepts is recognized here as well. The leading alternative keeps a literal "%%" from being
+     * treated as a placeholder.
+     *
+     * This is a copy and can therefore drift apart when sprintf-js changes - 1.1.x for example added
+     * the conversion characters "t", "T" and "v". The spec of this mixin compares the expression
+     * against the parser of the installed sprintf-js, so such a change fails the tests. When that
+     * happens, align this expression with the "placeholder" regex of the new version.
+     */
+    PLACEHOLDER_REGEX: /%%|%(?:([1-9]\d*)\$|\(([^)]+)\))?(\+)?(0|'[^$])?(-)?(\d+)?(?:\.(\d+))?([b-gijosuxX])/g,
```
ChristianMayer: Using a stateful regex in a static variable feels wrong.

ChristianMayer (changes requested):

source/class/cv/ui/common/BasicUpdate.js
joltcoke: Thanks! Addressed in 393ed46

ChristianMayer (approved):

PR #1489, opened 2026-09-08
Add a d3 based diagram plugin (diagram2)

joltcoke:
This adds `diagram2`, a successor of the flot based `diagram` plugin that draws with d3.

**Draft on purpose: this does not run yet.** The bundled `d3.min.js` is a slim build with 44 exports and has no `zoom`, no `drag`, no `axisRight` and no `curveStepAfter` - the plugin needs all four, for panning and zooming, for a value axis on the right and for step graphs. The entry file and the rollup setup of that build are not in the repository, so I could not simply extend it. I would like to agree on the way forward before finishing this:

- extend the slim build: its entry file would have to come back into the repository, `zoom` and the rest added, and the bundle rebuilt, or
- ship the official d3 distribution instead: a superset of the current exports, the file grows from 78 kB to 280 kB and is still only requested when a diagram or a chart tile is on the page.

I have both working locally and will push whichever you prefer into this PR.

### What it is

`diagram2` and `diagram2_info` read the same XML the `diagram` plugin reads - attributes, child elements and defaults are unchanged, only the widget names differ, so both plugins can live in the same configuration and be compared side by side. Data comes from the existing `cv.io.timeseries` sources (`rrd`, `influx`, `openhab`, `demo`).

The drawing follows the flot based plugin deliberately closely, because the aim is a replacement, not a new look: the same colour palette, flot's 2 % autoscale margin and its tick counts, the translucent legend background with the framed colour box, the 24 hour time format table, and the tooltip styling the designs already carry. Panning and zooming in the popup use d3-zoom with flot's zoom factor of 1.5.

Four attributes are new and all default to the current diagram2 look, so nothing changes without asking for it: `gridwidth`, `gridopacity`, `borderwidth` and `linewidth`. The migration guide names the values that reproduce the flot diagram (`borderwidth="2" linewidth="2" gridopacity="1"`).

### What is not in here

The installation this was written for reads its data from ioBroker, and that source is left out on purpose so this PR stays about the diagram. The plugin has a branch per source type, so adding it later is one case in `getTimeSeriesSource()`.

### Testing

With the full d3 build in place locally:

- 20 new specs in `source/test/karma/plugins/diagram2-spec.js`, the whole suite is green.
- Drawing checked against a demo configuration: one and two value axes, two axes on the same side, axis titles, `style` lines, bars and points, steps with and without fill, legend in all four corners, `previewlabels` on and off, tooltip, `refresh`, `popup` set and unset.
- Pan and zoom checked with mouse, wheel and double click, and on iOS in the simulator (iPhone 14, iOS 16.4) with one finger and two.
- The chart tile of the tile structure was checked after the library change.

On this branch, without the fuller d3 build: `qx compile`, `eslint` and the whole karma suite are green, the plugin just cannot draw yet.

### Known gaps

- `rrd`, `influx` and `openhab` are covered by unit tests, but I have no backend here to run them against.
- No screenshots for the manual yet, the pages use the generated attribute and element tables.


joltcoke:
@peuter I need your help here. The current bundled D3 version is stripped to the needs of the tiles component. Could you please add the for functions listed above? Or provide me the build scripts? Thanks!

peuter:
@joltcoke I added the build scripts that create the d3.min.js(+map) files to the develop branch.

You can modify the file `utils/d3/d3.js`  for your needs and run `npm run update-d3` to generate the libs.

